PR #65, opened 2023-06-06
Added actions workflow for building and running the education platform and its unit tests

barnettwilliam:
For issue #5, the unit test will be run on every commit to education platform and a report summarising the results available on github.


szschaler (approved):
LGTM

Issue #64, opened 2023-05-31
Running the platform: not working on Windows 10


arturboronat:
## Not able to run platform on Windows 10

Below I'm explaining the problems I experienced to install the platform on Windows 10. I explored several alternatives, and I'm reporting where I got stuck in each of them.

## Working with the sources

I tried to have a go at executing the platform on my computer, both using the sources and using the containerised version.

Find below the problems I found in both cases:

### Installation notes: to run education platform

Before `npm run build --workspaces`:

```
npm install -g tsc
npm install -g webpack
npm install typescript
``` 

Before running `npm run start --workspace=platform`, the docker daemon must be running and I executed this command from an elevated powershell terminal.

The tokenserver run but I could not access it: `http://127.0.0.1:10000/` gives `Cannot GET /`.

### Running activities

When running the Epsilon playground example given [here](https://github.com/mdenet/educationplatform#specifying-the-activities), the page hangs because I am not running anything on port 8082. The section on [external tool functions](https://github.com/mdenet/educationplatform#external-tool-functions) suggests that I need to run the dockerised epsilon playground.

#### From sources

When choosing to compile the image and run it locally I got this error:

```
WARNING: buildx: failed to read current commit information with git rev-parse --is-inside-work-tree
```

#### Containerised version

It executed well from the already compiled dockerised image. And I can run the Epsilon playground with its examples on port 8082.

### Executing Epsilon tasks from MDENet playground

Going back to the MDENet playground, when I execute `http://127.0.0.1:8082/epsilon-example/epsilon-example_activity.json` NGINX fails.

So when I execute `http://127.0.0.1:8080/?activities=http://127.0.0.1:8082/epsilon-example/epsilon-example_activity.json`, the platforms asks me for my GitHub credentials: 
* If I click on `Sign in with github`, nothing happens.  
* If I click on `Continue without signing in`, it hangs. 

### Problems found

Currently, I am stuck here with the following problens:
* When running from sources, the tokenserver is not running and I cannot access the example given.
* I am missing more instructions on how run the server with the examples so that `http://127.0.0.1:8082/epsilon-example/epsilon-example_activity.json` works. The examples project is [here](https://github.com/mdenet/educationplatform-examples/tree/main/epsilon-example). I don't see how to make the connection.

## Problem from Windows (Powershell)

On Windows, I needed `git config --global core.longpaths true` for cloning the repo.

[...]

PR #66, opened 2023-06-08
Added gitattributes to set line endings for windows build support

barnettwilliam:
plus updated readme with added windows specfic steps.

mdenet/educationplatform#64, adding gitattributes to all repositories so future commits are normalised properly across all platforms developers are using.


szschaler (approved):
LGTM

PR #72, opened 2023-07-18
71 support configuration files in yaml format

barnettwilliam:
Support for yaml configuration files in addition to json, they can be used interchangeably.


platform/src/ActivityManager.js
```diff
@@ -252,25 +251,9 @@ class ActivityManager {
      * Fetches the contents of the activity with the provided ID
      */ 
     fetchActivity(id) {
-        if (!this.hasActivity(id)) {
-            var xhr = new XMLHttpRequest();
```
szschaler, line 256: where is this fetching happening now?

platform/src/Utility.js
```diff
@@ -171,4 +173,32 @@ export function urlParamPrivateRepo(){
...
+}
+
+/**
+ * Parses the platform configuration files, YAML and JSON types are supported. 
+ * @param {String} contents the configuration file contents
+ * @param {String} extenstion the configuration file extenstion
+ * @returns the parsed configuration object
+ */
+export function parseConfigFile(contents, extenstion="yml"){
+
```
szschaler, line 185: Typo: "extenstion" -> "extension"

szschaler (changes requested):
Generally looks good, but have we lost the loading of activities? Also, one minor edit.

platform/src/Utility.js
barnettwilliam: Corrected in 393b3ec2bc292a452777745e3a254b8dcfba0a99

platform/src/ActivityManager.js
barnettwilliam: The removed code only loaded stored activities when following a previously generated Epsilon Playground link. So the activity loading behaviour for us is unchanged as we now use github as a mechanism for storing user changes.

The actual activities are loaded on Line 40 of platform/src/ActivityManager.js

szschaler (approved):

PR #83, opened 2023-09-20
Experimental languageworkbenchsupport

barnettwilliam:
Base language workbench functionality demonstrated with Xtext.


platform/src/Playground.js
```diff
@@ -14,8 +16,9 @@ import { ToolManager as ToolsManager } from './ToolsManager.js';
 import { ConsolePanel } from "./ConsolePanel.js";
 import { ProgramPanel } from "./ProgramPanel.js";
 import { OutputPanel } from "./OutputPanel.js";
-import { TestPanel } from './TestPanel .js';
+import { TestPanel } from './TestPanel.js';
 import { BlankPanel } from './BlankPanel .js';
+import { XtextEditorPanel } from './XtextEditorPanel.js';
```
szschaler, line 21: This feels quite specialised. Do we have to have an explicit reference to Xtext in the main platform code?

Note, I'm not asking for this to be changed before we merge the code, but we should consider whether we can further refactor at some point.

platform/src/Playground.js
```diff
@@ -240,6 +242,16 @@ function initialisePanels() {
                 
                 newPanel.hideEditor();
                 newPanel.showDiagram();
+            break;
+
+            case "XtextEditorPanel":
```
szschaler, line 247: See above...

platform/src/Playground.js
```diff
@@ -867,6 +900,39 @@ function savePanelContents(event){
...
+async function checkEditorReady(statusUrl, url, editorPanel, editorActivity){
+
+   let response  = await fetch(statusUrl);
+
+   if (response.status == 200){ 
+        const result = await response.json();
+
+        if (!result.editorReady){
+            await new Promise(resolve => setTimeout(resolve, 2000));
+            await checkEditorReady(statusUrl, url, editorPanel, editorActivity);
```
szschaler, line 919: Why are we doing this active waiting? Shouldn't this simply be an HTTP call that waits for a promise to get the result?

platform/src/Playground.js
```diff
@@ -867,6 +900,39 @@ function savePanelContents(event){
...
+
+        } else {
+            // Successful 
+            console.log("Editor ready.");
+            sessionStorage.setItem( editorPanel , url );
```
szschaler, line 924: That's very specific. It's OK for now, but it may make it difficult to do multiple languages or similar at some point in the future... It's definitely important to document this approach carefully that ties us to one editorPanel.

szschaler (approved):
Generally looks good, though I'm a bit concerned that the platform now contains Xtext-specific code.

platform/src/Playground.js
barnettwilliam: Yes there is work to be done generalise the Xtext workbench, I've added it as a task to the tracking issue #81. 

To use the generated editor that is an XtextServelet the education platform uses xtext-ace.js (provided by xtext) to handle and communicate changes to the editor instance. This is mostly because the platform uses the ace editor so it minimsed changes required to the platform. 

However, the LSP seems to be more widely supported and Xtext can also generate a LSP for languages created using it. The ace editor doesn't support LSP, however, there is support for it in the monaco editor/eclipse theia. Therefore, we should consider switching editors and using LSP for communication between generated editors.

platform/src/Playground.js
barnettwilliam: Agree that we should use server sent events to detect this rather than polling, added task to #81 .

platform/src/Playground.js
barnettwilliam: It is the id to the panel so it is not specifically an Xtext panel but there is an assumption that we will be able to generalise the XtextEditorPanel class into something for all language workbenches. Otherwise this will have to be revised. I'll update the parameter names for now.

barnettwilliam: Names updated in 16a9b13ece770c7d5f42407c71ab2bba858a2b63

szschaler (approved):

PR #86, opened 2023-09-27
Moved common utilities into the educationplatform static

barnettwilliam:
Moved common utilities into the educationplatform static. 

Updated npm build script to copy files to the dist folder to simplify deployment.

### DIFF
--- a/platform/public/common/utility.json
+++ b/platform/public/common/utility.json
@@ -1,0 +1,47 @@
+{
+    "tool": 
+        {
+            "id": "utility",
+    
+            "name": "Common Utilities",
+    
+            "version": "0.0.1",
+            
+            "author": "MDENet",
+            
+            "homepage": "",
+    
+            "functions": [    
+            ],
+    
+    
+    
+            "panelDefs": [
+    
+                {
+    
+                    "id": "console",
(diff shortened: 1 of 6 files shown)


platform/public/common/utility.json
```diff
@@ -0,0 +1,47 @@
...
+    
+                {
+    
+                    "id": "console",
+    
+                    "name": "console",
+    
+                    "panelclass": "ConsolePanel",
+    
+                    "icon": "images/console.gif"
```
szschaler, line 29: Where does this image file live?

szschaler (approved):
Would be good to cross-reference https://github.com/mdenet/platformtools/pull/18 from this PR for documentation

platform/public/common/utility.json
barnettwilliam: @szschaler  when direct the path used it is relative to the playground base url so it'll be http://127.0.0.1/8080/images/console.gif.

The other way of referencing icons is by using the css id which are loaded relative to the location of the css e.g. the tool static urls.

szschaler: Thanks for clarifying.

PR #90, opened 2023-10-13
Fixed broken navigation menu links for private activities

barnettwilliam:
Added privaterepo parameter to menu links when a private activity is loaded and also cleaned authentication state and code parameters from url following authentication.


barnettwilliam:
Raised #91 for detecting private repositories and removal `privaterepo` url parameter.

szschaler (approved):

PR #95, opened 2023-10-31
Hide the login box by default and only show for private repo

barnettwilliam:
Hide the login box by default and only show it when a private repo is given. This prevents the login box's brief display when loading a private repo mdenet/educationplatform-docker#2.

### DIFF
--- a/platform/src/Playground.js
+++ b/platform/src/Playground.js
@@ -60,6 +60,8 @@ if (!urlParamPrivateRepo()){
     // Public repo so no need to authenticate
     initialiseActivity();
     PlaygroundUtility.hideLogin();
+} else {
+    PlaygroundUtility.showLogin();
 }
 
 document.getElementById("btnlogin").onclick= async () => {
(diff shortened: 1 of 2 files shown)


platform/src/Playground.js
```diff
@@ -60,6 +60,8 @@ if (!urlParamPrivateRepo()){
     // Public repo so no need to authenticate
     initialiseActivity();
     PlaygroundUtility.hideLogin();
```
szschaler, line 62: Is this line still needed?

platform/src/Playground.js
```diff
@@ -60,6 +60,8 @@ if (!urlParamPrivateRepo()){
...
+} else {
+    PlaygroundUtility.showLogin();
```
szschaler, line 64: This sets visibility to `block` when it was `inline` originally. Does it matter?

szschaler (approved):
Generally OK, though I have two small questions.

platform/src/Playground.js
barnettwilliam: Agreed this can be removed.

platform/src/Playground.js
barnettwilliam: It affects the margins but and unnoticeable in this case as it's a floating box

szschaler (approved):

PR #100, opened 2023-11-01
Add panel definition for generated code

agarciadom:
The standard panel definitions only have one `OutputPanel`, and it seems to be for a "Problems" view (e.g. for EVL or OCL). It'd be good to have a panel definition for generated code, as that scenario could be quite common.

### DIFF
--- a/platform/public/common/utility.json
+++ b/platform/public/common/utility.json
@@ -40,7 +40,20 @@
     
                     "icon": "problems"
     
+                },
+
+                {
+    
+                    "id": "code",
+    
+                    "name": "code",
+                    
+                    "panelclass": "OutputPanel",
+    
+                    "icon": "editor"
+    
                 }
+
             ]
         }
     } 


szschaler:
@agarciadom thanks. Where is this panel used? Do we really need a separate **type** of panel? 

The platform idea would be that there would be a suitable panel for the type of contents to show. For example, in a M2M scenario, you would have a source model, perhaps expresses in FLEXMI, and a transformation specification, say in ETL. You would then have a FlexMI panel for the source model, an XMI panel (or something else) for the meta-model, and an ETL panel for the transformation specification. You would _also_ have a second FLEXMI panel for the target model, which might initially be empty, but can also be linked to a file in the underlying repository.

You would then define the action parameters in the activity specification such that it places its output into the second FLEXMI panel. The platform should take care to do any necessary conversions between formats, assuming that appropriate conversion functions are made available by the tools (for example, the FLEXMI tool has a conversion function from and two XMI).

What we _don't_ currently support yet is actions that produce multiple outputs. I think there is an issue for that already, but @barnettwilliam will correct me if I'm wrong...

agarciadom:
The main thing is that if you run an EGL transformation, there is no predefined panel to put in the output with an icon that indicates it's generated text. The existing `OutputPanel` forces you to use a "Problems" icon. We could extend the Epsilon tool to put in a panel definition there, but I think that it's general enough (any M2T transformation would need it) that it should be part of the predefined ones.

(Basically, I need this panel for EGL.)

I'm not looking at showing the results of a M2M transformation in Flexmi format at the moment - my idea was to just show the produced model as a PlantUML diagram, just like what we do already in the Epsilon Playground.

szschaler:
I see. Would it then make more sense to have a plantUML panel that can show either the text or the diagram and use that? Basically, I'm hesitant about introducing a new centrally hard-coded panel type if we can avoid it as it makes the main platform more complex. Instead, I would prefer having this be supported by an appropriate tool...

barnettwilliam:
The OutputPanel class has similar functionality to the plantUML panel you describe @szschaler  which is derived from ModelPanel. 

OutputPanel extends ModelPanel to include support for displaying multiple files.

Perhaps its just a case of being able to override the icons in the activity panel definitions? I think this is something we have talked about but it's not implemented.

szschaler:
I see. My bad, I didn't realise that this was exactly the functionality already. How does this fit in with our mechanisms for processing outputs from action functions in a generic manner?

barnettwilliam:
I don't think the ModelPanel is directly using plant uml for the model to diagram conversion but the playground was using something like kroki service and egl.

The `panel id` to output the result of a tool function when a button is clicked is specified in the activity's configuration under actions and the class of panel of the output is declared in the activity's config file under panels.  The handling is just display the result in the output panel so the panel the result is displayed in must support result's type.  Tool services define panels that display specific types e.g. emfatic, flexmi. 

There is an issue to add conversion of toolfunction output type to the destination panel which should be similar to the type conversion we do for tool function input parameters. I don't see any difference other than a panel may be able to display multiple types. I need to look into the details of the types and  how they are being visualised to understand exactly what is supported by the ModelPanel.

agarciadom:
> Perhaps its just a case of being able to override the icons in the activity panel definitions? I think this is something we have talked about but it's not implemented.

Yes, maybe this would be a better choice, i.e. being able to just change the icon for a given panel.

szschaler:
This seems to have got stuck. @barnettwilliam, can we please discuss it on Wednesday to decide what to do with it?

barnettwilliam:
@agarciadom support for panel customisation for activity files has now been added by the fix for #102 by declaring an `icon` in the panel https://github.com/mdenet/educationplatform/wiki/Creating-an-Activity#panel so you should no longer need to define a new panel type in this case. Could you confirm this resolves your issue?

agarciadom:
If the argument is that we don't want to have multiple standard `OutputPanel` panel definitions in the platform when it's possible to just customise the icon, then it would make more sense to call the generic OutputPanel `output` and not `problems` (which seems to be a more specific case). In any case, I'm happy to close the PR - I'll resolve this from the Epsilon side.

Issue #98, opened 2023-11-01
"Executing program..." popup does not automatically disappear


agarciadom:
(Moved here from https://github.com/mdenet/educationplatform-docker/issues/28)

I tried running the example activity with EOL, ETL, and EVL:

http://127.0.0.1:8080/?activities=https%3A%2F%2Fraw.githubusercontent.com%2Fmdenet%2Feducationplatform-examples%2Fmain%2Fepsilon-example%2Fepsilon-example_activity.yml

I pressed the Run button, and saw the "Executing program..." popup appear:

![image](https://github.com/mdenet/educationplatform-docker/assets/46504/17f9a827-9cee-4318-af19-04fdc7b979b1)

I could see the output on the Console very quickly, but the "Executing program..." popup did not automatically disappear, and it would not be obvious to a first-time user that you can click on it to dismiss it.

Would it be possible to make it automatically disappear, or at least have an "X" icon in a corner so it's obvious you can dismiss it?

PR #103, opened 2023-11-07
Clear notifications for activities on handling a response and added icon

barnettwilliam:
Clear notifications for activities on handling a response and added cross icon to notifications to indicate that they can be closed.

To fix - https://github.com/mdenet/educationplatform/issues/98

notification_with_cross


szschaler (approved):
LGTM. I wonder whether there is an opportunity to refactor those three notification methods to DRY up the code a bit, but it's not a big deal...

PR #105, opened 2023-11-08
YAMTL playground: metro4, url encoding, puml

arturboronat:
## Context

To facilitate the publication of the bundled platform as a static resource akin to the approach used in the Epsilon Playground, I have introduced several changes. These modifications enable the platform to run without the necessity of Docker.

The objective is to incorporate these updates in a way that retains the functionality of the existing Docker deployment.

## Proposed Changes

- The `import metro4` statement does not function in this static context - at least for me.
- Correct URL encoding is essential; otherwise, characters such as `:` and `/` are replaced with their percent-encoding. While the platform works, the url cannot copy/pasted directly anymore.
- A probable issue is detected with `newPanelId` within the `OutputConsole`, where it erroneously references the panel's ID instead of the expected `PanelDef` ID.
- The `outputType` currently set as a constant must be dynamically sourced from the `action` object's `outputType` property to ensure the correct rendering of PlantUML diagrams.

### DIFF
--- a/platform/src/Playground.js
+++ b/platform/src/Playground.js
@@ -7,7 +7,11 @@ import 'ace-builds/src-min-noconflict/mode-html';
 import 'ace-builds/src-min-noconflict/ext-modelist';
 import 'ace-builds/src-min-noconflict/ext-language_tools';
 
-import 'metro4';
+// BEGIN YAMTL Playground
+// FIXME: this is ad-hoc code
+// import 'metro4' // this import does not work for local debugging
+import '../../node_modules/metro4/build/metro.js' // this import does not work for docker
+// END YAMTL Playground
 
 import { FileHandler } from './FileHandler.js';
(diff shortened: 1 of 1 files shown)


barnettwilliam:
Thank you for these Artur I will look into them.  The metro4 import issue points to something with build configuration as you shouldn't have to directly reference a full `node_module` path.

arturboronat:
Thanks. I'm using the configuration in https://github.com/yamtl/educationplatform/tree/yamtl-playground-08112023 which comes from the original repo. Not sure where the problem may be.

szschaler:
@barnettwilliam, can we discuss this when we meet this week?

barnettwilliam:
Thank you for these Artur, I have opened issue #108 for the metro import issue and moved url encoding and puml changes to pull request #109 therefore closing this pull request.

Issue #102, opened 2023-11-07
Add support for panel customisation in activity files


barnettwilliam:
It should be possible to customise panels defined in the an activity overriding the defaults defined in tool configurations panel definitions.  

For example panel icons noted by https://github.com/mdenet/educationplatform/pull/100

PR #106, opened 2023-11-10
Support for panel customisation in activity files

barnettwilliam:
Added support for customising panel icon and buttons from the activity configuration for #102 and added Button class to improve cohesion.

Updated documentation located mdenet/educationplatform.wiki@e788820f37e2715d2c703504779cf3560983c56b
on branch `102-add-support-for-panel-customisation-in-activity-files`

### DIFF
--- a/platform/src/ConsolePanel.js
+++ b/platform/src/ConsolePanel.js
@@ -8,21 +8,20 @@ class ConsolePanel extends Panel {
         super(id);
         this.editor.setReadOnly(true);
         this.editor.setValue("", 1);
-        this.element.dataset.customButtons = JSON.stringify(this.getButtons());
+
+        let buttons = [];
+        let clearButton = new Button(
+            { id:"clear", 
+              hint:"Clear the console", 
+              internal: `panels.find((p) => p.id==="${this.id}").editor.setValue('')`,
+              icon: "clear" }, 
+            this.id
+        );
+        buttons.push(clearButton); 
+        this.addButtons(buttons);
+
         this.detectHyperlinks(this.editor);
         this.setTitleAndIcon("Console", "console");   
-    }
-
-    getButtons() {
-        let clearButton = new Button({
-                id:"clear", 
-                hint:"Clear the console", 
-                internal: `panels.find((p) => p.id==="${this.id}").editor.setValue('')`,
-                icon: "clear" 
-            }, this.id
-        );
-    
-        return [clearButton.getView()];
     }
 
     setOutput(str) {
--- a/platform/src/ConsolePanel.js
+++ b/platform/src/ConsolePanel.js
@@ -1,5 +1,6 @@
 import { Panel } from "./Panel.js";
 import { define } from "ace-builds";
+import { Button } from "./Button.js";
 
 class ConsolePanel extends Panel {
 
@@ -13,11 +14,15 @@ class ConsolePanel extends Panel {
     }
 
     getButtons() {
-        return [{
-            html: this.buttonHtml("clear", "Clear the console"),
-            cls: "sys-button",
-            onclick: "consolePanel.setValue('')"
-        }];
+        let clearButton = new Button({
+                id:"clear", 
+                hint:"Clear the console", 
+                internal: `panels.find((p) => p.id==="${this.id}").editor.setValue('')`,
+                icon: "clear" 
+            }, this.id
+        );
+    
+        return [clearButton.getView()];
     }
 
     setOutput(str) {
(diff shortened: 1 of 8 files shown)


platform/src/OutputPanel.js
```diff
@@ -1,5 +1,7 @@
+
 import { ModelPanel } from "./ModelPanel.js";
-import { language } from "./Playground.js";
+import { language } from "./Playground.js"
```
szschaler, line 3: Why did you remove the semicolon here?

platform/src/Panel.js
```diff
@@ -97,27 +97,14 @@ class Panel {
...
 
         var buttonViewData= buttons.map( (btn) => {
-            var buttonData={};
-            
-            buttonData.html= this.buttonHtml(btn.icon, btn.hint);
-            buttonData.cls= "sys-button";
-            buttonData.onclick= btn.action;
-
-            return buttonData;
+            return btn.getView();
```
szschaler, line 107: But elsewhere you're already calling `.getView()` when defining the buttons. Are these consistent?

platform/src/Playground.js
```diff
@@ -262,8 +263,8 @@ function initialisePanels() {
    /**
      * Create a panel for a given panel config entry
      * 
-     * @param {string} panel 
-     * @return {Panel}
+     * @param {Object} activity config panel definition 
+     * @return {Panel} platform Panel
      */
     function createPanelForDefinitionId(panel){
```
szschaler, line 269: Is the parameter name `panel` still correct?

platform/src/ConsolePanel.js
```diff
@@ -13,11 +14,15 @@ class ConsolePanel extends Panel {
     }
 
     getButtons() {
```
szschaler, line 16: Where is this method used?

szschaler (commented):
Generally looks good except for some bits that I didn't understand...

platform/src/ConsolePanel.js
barnettwilliam: It's called in the ConsolePanel constructor on line 11 which sets the member variable that gets used to render the buttons.

szschaler: Panel defines an `addButtons` method. Why is that not used in the `ConsolePanel` constructor?

szschaler:
I guess what confuses me is that we seem to have three different mechanisms for defining buttons with this PR. They appear to all be subtly different, but I may also just be misinterpreting...

platform/src/Playground.js
barnettwilliam: The parameter name is as intended the updated description removed the parameter name.
I'll update it to `@param {Object} panel -  The activity config panel definition.`

The jsdoc comment style supports dashes to separate paramter name from description which make this clearer raised #107 to apply consistently.

platform/src/Panel.js
barnettwilliam: There are two ways Buttons are intialised
- for built-in panels that aren't customisable they are intialised in the panel constructor with a call to the panel's getButtons() method e.g. the ConsolePanel line 11.

- for panels that are customisable like those defined by tools in the configuration they use the Panel super class'  addButtons() method called during platform intialisation in Playground.js line 352.

With the added customisation of panels that override those defined in tool panel def there are likely to be odd behaviours introduced e.g. overriding a non-customisable panel's button.

platform/src/OutputPanel.js
barnettwilliam: Agree this was unintentional so will revert.

platform/src/Panel.js
szschaler: But why are those uses not consistent? One uses `getView()` in the definition of the button, the other uses it when adding the button via `addButtons()`.

How would buttons be overwritten and what would happen in that case?

barnettwilliam: I agree that the addButton() can be used in same way for the built-in panels so the getButtons() can be easily removed.

For overwriting the built-in panels buttons its likely you would be able to define buttons new buttons but not reference the panel defined ones by id. I will test this case though to see.

barnettwilliam: For the case of trying to overwrite a built-in panel's buttons the the panels config would have no buttons declared since they are instantiated by the panel's js. Therefore panelDefinition.buttons is undefined so the buttons are not modified by lines 335 or 352 - this is the intended behaviour confirmed by test.

barnettwilliam:
You were right to be confused, I have added the mentioned fixes in 1032f28d23a6bfd62a52818ba38a225ddbc0cf43

PR #118, opened 2023-11-26
115 add validation for config references

barnettwilliam:
Added check for: activity panel ID refs exist, activity button refs exist, and activity panel ref to tool panel. With some minor null checks added so that the error information isn't lost for display.

### DIFF
--- a/platform/src/ActivityManager.js
+++ b/platform/src/ActivityManager.js
@@ -338,7 +338,12 @@ class ActivityManager {
 
                 // Resolve the panel definition reference  
                 if ( typeof apanel.ref == "string" ){
-                    apanel.ref = this.accessPanelDef(apanel.ref);
+                    
+                    const panelDef = this.accessPanelDef(apanel.ref);
+
+                    if (panelDef != null){
+                        apanel.ref = panelDef;
+                    }
                 }
             }
         
@@ -443,7 +448,7 @@ class ActivityManager {
 
         for (const panel of  this.activities[activityId].panels) {
             
-            if (this.accessPanelDef(panel.ref).generated  && !this.isPanelGenerated(panel.id) ){
+            if (this.accessPanelDef(panel.ref)?.generated  && !this.isPanelGenerated(panel.id) ){
                 generatedPanelFound = this.accessPanelDef(panel.ref).generated;
                 break;
             }
--- a/platform/src/ActivityValidator.js
+++ b/platform/src/ActivityValidator.js
@@ -1,4 +1,4 @@
-import { ConfigValidationError } from './ConfigValidationError';
+import { ConfigValidationError, CONTEXT_SEPARATOR } from './ConfigValidationError';
 
 const ERROR_CATEGORY = "ERROR_CONFIG_DEF"
 
@@ -29,7 +29,7 @@ class ActivityValidator {
             if (!this.idExists(activity.panels, id)){
                     errors.push( 
                         new ConfigValidationError(ERROR_CATEGORY, "A panel does not exist for the given id.",  
-                        `${activity.id} -> layout -> ${id}`, errorFileType.ACTIVITY)
+                        `activity.id: ${activity.id}, layout ${CONTEXT_SEPARATOR} area ${CONTEXT_SEPARATOR} ${id}`, errorFileType.ACTIVITY)
                 );
             }
         });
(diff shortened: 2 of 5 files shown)


platform/src/ActivityManager.js
```diff
@@ -338,7 +338,12 @@ class ActivityManager {
 
                 // Resolve the panel definition reference  
                 if ( typeof apanel.ref == "string" ){
-                    apanel.ref = this.accessPanelDef(apanel.ref);
+                    
+                    const panelDef = this.accessPanelDef(apanel.ref);
+
+                    if (panelDef != null){
+                        apanel.ref = panelDef;
+                    }
```
szschaler, line 346: Should you not report an error in this case?

platform/src/ActivityValidator.js
```diff
@@ -0,0 +1,208 @@
...
+     * @returns {ConfigValidationError[]} Array of configuration errors, empty if there are none.
+     */
+    static checkActionIdsExist(action, panels){
+        let errors = [];
+
+        if (!this.idExists( panels, action.source.id)){ // action ids are resolved so get id from object
+            errors.push( 
+                new ConfigValidationError(ERROR_CATEGORY, "A panel does not exist for the given id.",  
+                `actions -> source: ${action.source}`, errorFileType.ACTIVITY)
+            );
```
szschaler, line 68: Not `${action.source.id}`?

platform/src/ActivityValidator.js
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        if (!this.idExists( panels, action.output.id)){ // action ids are resolved so get id from object
+            errors.push( 
+                new ConfigValidationError(ERROR_CATEGORY, "A panel does not exist for the given id.",  
+                `actions -> output: ${action.output}`, errorFileType.ACTIVITY)
```
szschaler, line 74: Not `${action.output.id}`?

platform/src/ActivityValidator.js
szschaler: And how does the error message identify the specific action?

platform/src/ActivityValidator.js
```diff
@@ -0,0 +1,208 @@
...
+            );
+        }
+
+        if ( (action.outputConsole !=null) && !this.idExists( panels, action.outputConsole.id)){ // action ids are resolved so get id from object
+            errors.push( 
+                new ConfigValidationError(ERROR_CATEGORY, "A panel does not exist for the given id.",  
+                `actions -> outputConsole: ${action.outputConsole}`, errorFileType.ACTIVITY)
```
szschaler, line 81: Not `${action.outputConsole.id`?

szschaler (commented):
Generally looks good, some minor questions...

platform/src/ActivityManager.js
barnettwilliam: This is reported by checkPanelRefs() from the ActivityValidator, I'm trying to keep all the validation checks together to make reuse easier in the vscode extension.

platform/src/ActivityValidator.js
barnettwilliam: As per https://github.com/mdenet/educationplatform/pull/118/files/dcc62b0aecc8468d790dbaba55466bc7792aab05#r1406005085

platform/src/ActivityValidator.js
barnettwilliam: As per https://github.com/mdenet/educationplatform/pull/118/files/dcc62b0aecc8468d790dbaba55466bc7792aab05#r1406005085

platform/src/ActivityValidator.js
szschaler: Hmm, but on Line 64 just a little bit above, you are using `action.source.id`. Wouldn't this be undefined in that place in that case?

barnettwilliam: Yes the idExists() function that's called on line 64 returns false if an null/undefined id is given, maybe this obscures the validation condition. The null check could be made the condition and a comment to note why to make it clearer.

barnettwilliam: The null check of the resolved panel id e.g. `action.source.id` is all that is needed to determine whether the panels id exists for the `checkActionIdsExist()` checks because `action.source` etc have been resolved earlier from a panel id and replaced with the panel object. So an alternative check that is equivalent is checking whether `action.source` is an object. 

The idea behind using `idExists()` is to double-check that the id of a resolved panel exists within the activities' panels for robustness.

szschaler: I would suggest making this explicit by doing both checks: `action.source isObject && checkActionIdsExist(...`. WDYT?

barnettwilliam: Yes I agree and will update.

barnettwilliam: Fixed in 40616a5525ed9b6fb0d7b2dc4de5fc691599691c

platform/src/ActivityValidator.js
barnettwilliam: Fixed in 40616a5525ed9b6fb0d7b2dc4de5fc691599691c

barnettwilliam:
Related pr https://github.com/mdenet/platformtools/pull/20

szschaler (approved):
Thanks for these changes. Much clearer now.

Issue #113, opened 2023-11-21
Add semantic validation of config files


barnettwilliam:

barnettwilliam:
Separated groups of validation checks into their own issues to be incrementally completed.

szschaler:
We will put this on the back burner for now as a potential student project

PR #119, opened 2023-11-26
114 add informative error messages eg when a required tool is missing etc

barnettwilliam:
Added exception handling on failure to access tool and activity configurations and display appropriate error message on failure. More detailed error messages to be added as part of #113.

### DIFF
--- a/platform/src/ToolsManager.js
+++ b/platform/src/ToolsManager.js
@@ -2,6 +2,7 @@ import { parseConfigFile } from "./Utility.js";
 import { FunctionRegistry } from "../src/FunctionRegistry.js"
 import { ActionFunction } from "./ActionFunction.js";
 import { ToolConfigValidator } from "./ToolConfigValidator.js";
+import { EducationPlatformError } from "./EducationPlatformError.js";
 
 class ToolManager {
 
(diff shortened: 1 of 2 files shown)


platform/src/ToolsManager.js
```diff
@@ -52,8 +53,17 @@ class ToolManager {
...
-    
+
+            try{
+                xhr.send();
+            } catch (e) {
+                if (e instanceof DOMException){
+                    errors.push( new EducationPlatformError(`A tool configuration file was not accessible at: ${toolUrl.url}. 
+                                                            Check the tool's url given in the activity file is correct and the tool 
+                                                            service is still available.`) );
+                }
```
szschaler, line 64: Would it make sense to `return errors;` here?

szschaler (commented):
Other than the one suggestion/query, lgtm.

platform/src/ToolsManager.js
barnettwilliam: The suggested change would mean only the first error is displayed on a tool being unreachable when there could be multiple tools that are unreachable. Currently all the unreachable tools are displayed so the teacher will see all the errors at once rather than re-run and then see the next etc. I prefer this behaviour but it could be changed if we want to incrementally show the errors.

szschaler: Understood. I was mainly worried about errors happening in the next few steps because they would still be executed even if the tool configuration couldn't be accessed.

szschaler (approved):

Issue #26, opened 2023-04-20
Lower entry barrier for first-time users


kolovos:
After cloning the repository and running `launch.sh`, `http://localhost:8080` produces a "No activity configuration has been specified." error message. A default activity configuration would probably be preferable for first-time users.

Visiting `http://localhost:8080/?activities=http://localhost:8082/epsilon-example/epsilon-example_activity.json` as instructed in the readme takes the user to a small list of examples, however, there are no icons next to the titles of the examples and running the examples does nothing. To fix the latter, I had to run the Epsilon Playground container on port `9000` as instructed later in the readme using `docker run -p 9000:80 eclipseepsilon/playground:1.6`. To help first-time users, I'd suggest either adding this command to the education platform's readme file or starting the Epsilon Playground container within the Docker compose file.

To make it even easier for the community to engage with the platform and provide early feedback, it'd be great if we could actually run a public instance on AWS/GCP (with no guarantees of correctness/stability of course).

szschaler:
We now have https://github.com/mdenet/educationplatform-docker, which is a repository from which an example configuration can be spun up comparatively simply. Setting up a public running instance remains open.

agarciadom:
I just tried the Docker Compose bit, and it looks good but I noticed that it flashes the Github auth for a second before it goes into the proper web UI, which could confuse a first-time user.

arturboronat:
@barnettwilliam  This seems a natural continuation of this issue: 

I've noticed that I often spend some time understanding the configuration of port wirings.

Specifically, for the epsilon tool, could you help me identify where 8082 is defined in the link: http://127.0.0.1:8080/?activities=http://127.0.0.1:8082/epsilon-example/epsilon-example_activity.json

I was wondering if this is documented on our wiki. If it is, could you kindly point me in the right direction? I was looking here, but wasn't certain: https://github.com/mdenet/educationplatform/wiki/Creating-an-Activity#resources

szschaler:
Given that this comes from shipping examples with the platform rather than as a separate standalone public repository, would it be easier to remove that option completely and only support activities that are hosted in GitHub repositories?

arturboronat:
The comment also applies to examples that are deployed in separate GitHub repos. Users need to know where to configure that port.

barnettwilliam:
The best place to find out is in the docker compose file in the top-level docker repository /educationplatform-docker/docker-compose.yml as defines the demo instance containers.

[...]

PR #120, opened 2023-11-27
Tokenserver removed unnecessary username request and log

barnettwilliam:
Removed unnecessary username request and log from the tokenserver and replaced with the current date and time.

In preparation for creating a publicly available instance #26

### DIFF
--- a/tokenserver/src/server.js
+++ b/tokenserver/src/server.js
@@ -51,8 +51,10 @@ expressApp.use(errorHandlingMiddleware);
 
 
 githubApp.on("token", async ({ token, octokit, expiresAt }) => {
-  const { data } = await octokit.request("GET /user");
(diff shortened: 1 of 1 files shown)


tokenserver/src/server.js
```diff
@@ -51,8 +51,10 @@ expressApp.use(errorHandlingMiddleware);
 
 
 githubApp.on("token", async ({ token, octokit, expiresAt }) => {
-  const { data } = await octokit.request("GET /user");
```
szschaler, line 54: Don't you need to keep this?

barnettwilliam: No this just gets the username which was printed as debug information it's not required as GitHub uses the token to authenticate the user.

szschaler: Ah, I see.

PR #125, opened 2023-11-28
Add support for customising the URL to the token server

agarciadom:
I have started to look into deploying a public instance of the platform. From what I can see, it seems like the token server URL was hardcoded into the JavaScript code:

```js
const TOKEN_HANDLER_URL = "http://127.0.0.1:10000";
```

This pull request allows for customising the URL of the token handler, by using a Webpack `DefinePlugin`. The line can be changed to:

```js
const TOKEN_HANDLER_URL = TOKEN_SERVER_URL || "http://127.0.0.1:10000";
```

The TOKEN_SERVER_URL value will be replaced on the fly by Webpack using the DefinePlugin. This would allow us to tweak the URL by using the appropriate value when calling `npm`, like this:

```shell
npm run build -- --env tokenServerUrl=http://foo:1234
```

I had to separate the `copyfiles` invocation into its own `npm run package` script, as otherwise I wasn't able to pass Webpack environment variables to the build.

### DIFF
--- a/platform/Dockerfile
+++ b/platform/Dockerfile
@@ -12,10 +12,9 @@ COPY package*.json ./
 
 COPY . .
 
-RUN npm install; npm run build
+ARG TOKEN_SERVER_URL=http://127.0.0.1:10000
 
-
-
+RUN npm install && npm run build -- --env tokenServerUrl=${TOKEN_SERVER_URL} && npm run package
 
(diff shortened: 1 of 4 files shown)


barnettwilliam (approved):
Looks good and tested changing the tokenserver via docker arg successfully.

PR #136, opened 2023-12-03
Only show 'Save' button when authenticated

### LINKED ISSUE #130: Save button should be hidden if saving isn't possible
The save button should not be displayed if saving is not possible when the user is not authenticated and doesn't have write access to the repository.

agarciadom:
Fixes #130 

This pull request adds the CSS class `hidden` (whose display value is now `!important`) to the Save button, and tweaks the code that handles the event when we authenticate to GitHub to remove that class, so the Save button is shown.

### DIFF
--- a/platform/public/css/custom.css
+++ b/platform/public/css/custom.css
@@ -165,7 +165,7 @@ body {
 }
 
 .hidden {
-	display:none;
+	display: none !important;
 }
 
 .ace_gutter {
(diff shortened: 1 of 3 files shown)


barnettwilliam (approved):
LGTM, tested and working with mdenet/educationplatform-docker/4d6fcd4e8d408db52d9b51877ac21e0043311785

barnettwilliam:
Along with mdenet/educationplatform-docker#38 means that the index.html in the the mdenet/educationplatform-docker repo no longer needs to be updated separately.

PR #137, opened 2023-12-03
Allow trusted origins to be configured via .env

agarciadom:
This should help with the customisability of the token server, by adding support for specifying the trusted origins by setting `TRUSTED_ORIGINS` to a comma-separated list of origins in the `.env` file.


barnettwilliam:
Looks good, these can apply to other servers that support the platform so agree with the name used and for single origins the express config doesn't require the array as originally used.

PR #138, opened 2023-12-03
Add 'activitiesFromReferrer' query string key

agarciadom:
I have tried using Github Classroom to set up an assignment based on a template repository, with the MDENet EP app installed into the classroom's organisation, and this has worked fine.

However, it requires students to manually tweak the `activities` query parameter, which I think is a bit unwieldy. In this PR, I add an alternative which uses the `Referrer` header sent by the browser when clicking on a link from the repository's `README` file. You would be able to add a link like this to your README:

```
protocol://host/?activitiesFromReferrer=/main/activity.json&privaterepo=true
```

This link would redirect to the `activity.json` file from the `main` branch of the repository whose README link we just clicked on, meaning we could set up a template repository and it would always work, so long as the private repo has the appropriate Github app installed.

This PR is pending to be tested in our current deployment - Cloudflare is taking a while to revalidate their cache.

### DIFF
--- a/platform/src/Playground.js
+++ b/platform/src/Playground.js
@@ -52,7 +52,38 @@ export var toolsManager;
 
 var urlParameters = new URLSearchParams(window.location.search);    
 
+// Redirect to activities file contained in referrer.
+//
+// This can be useful for providing a link to the MDENet EP from the README,
+// in a way that translates automatically to forks. It only works if the
+// referrer is something like https://github.com/user/repo, and would be used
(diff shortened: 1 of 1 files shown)


agarciadom:
Nevermind - this only works for README links from public repositories, apparently. When you click on a README link from a private repository, the Referer header only says `https://github.com`, so this wouldn't work. Closing as ineffective.

I'm actually not sure what we could do to make it so users can just follow a link - it doesn't look like there is any way to generate a link that references the fork details.

PR #148, opened 2023-12-08
Manage editor state when changing contents

### LINKED ISSUE #146: Avoid empty commits when saving
Related to #140 

Currently, saving creates a commit for each saveable panel, regardless of whether anything has actually changed. As a result, we are creating empty commits, which can be confusing to learners. We should only create commits that include the actually changed files and not create a commit at all if there is nothing to save.

It should be possible to use ACE's `isClean`/`markClean` functions in the `UndoManager` for this.

szschaler:
Closes #146

### DIFF
--- a/platform/src/ConsolePanel.js
+++ b/platform/src/ConsolePanel.js
@@ -33,6 +33,8 @@ class ConsolePanel extends Panel {
         element.style.color = "black";
         this.editor.getSession().setUseWrapMode(false);
         this.editor.setValue(str, 1);
+        // Reset undo manager
+        this.editor.session.getUndoManager().reset();
         // Scroll to the bottom. This works because we're really still using an ACE editor even in the console panel.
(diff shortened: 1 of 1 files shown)


platform/src/Panel.js
```diff
@@ -86,6 +92,8 @@ class Panel {
 
     setValue(value) {
         this.editor.setValue((value+""), 1);
+        // Reset undo manager
+        this.editor.session.getUndoManager().markClean();
```
barnettwilliam, line 96: This could be made to reset() rather than markclean() so that undo doesn't allow users to undo back to a blank panel as setValue is only used by the platform to populate the contents. Alternatively, we could add a method for resetting the undos to allow the platform to manage this is we want finer-grained control of this.

barnettwilliam (approved):
Generally looks good and prevents empty commits. It would be be nice to have visual feedback e.g. disable the save button and prevent undoing back to a blank panel but will raise separate issues for these.

PR #153, opened 2023-12-18
Use workspace top-level package lock

barnettwilliam:
When looking to resolve the flagged security issues, the platform and tokenserver should use the top-level workspace package lock rather than individually generated workspace-level package lock so that overrides can be used to ensure dependencies versions are configurable.

Updated the platform and tokenserver to use the top-level package-lock and removed unused workspace-level package-lock.


barnettwilliam:
Requires changes made in mdenet/educationplatform-docker#41

szschaler (approved):

PR #156, opened 2024-01-05
Add support for URL interpolation when setting up panels

szschaler:
This PR adds support for interpolation in panel URLs in the activity specification.

This is useful to allow dependent activities to access information from other activities. The concrete motivating example was the generation of Xtext editors, where I may wish to allow students to access the meta-model generated by Xtext in the activity where they can try out their new language. The URL to that file is generated by the platform (specifically, the xtext tool) and so isn't known at the time when the activity specification file is written.

With this PR, I can now write something like:

```
{
  "id": "panel-mm",
  "name": "The meta-model generated by Xtext",
  "ref": "ecore",
  "file": "{{ID-panel-turtles}}/xtext-resources/generated/meta-model.ecore"
},
```
and have `{{ID-panel-turtles}}` replaced from session storage associated with the `panel-turtles` panel at runtime. For generated editors, the platform stores the editor URL in that session storage key.

This is a somewhat pragmatic change, which could be made more powerful, but for now I suggest we keep things simple. We may need to invest in a more systematic architecture should there be a need for more flexibility at any point down the line.

The change consists of two separate changes:

1. Changing `ActivityManager.js` to perform the URL interpolation, and
2. Changing `FileHandler.js` to gracefully manage unknown types of URLs even for private repositories.

The latter change isn't strictly related to URL interpolation, but is required so that the generated URL can actually be used by the platform even when the activity lives in a private repository. To make this compatible with how file saving is managed, I have also adjusted the code for `canSave` in various panel types to ensure we only attempt to save a panel when it actually has an associated SHA and, thus, we know that it came from the original repository. This may make it difficult for the platform to save back to _public_ repositories.

### DIFF
--- a/platform/src/ActivityManager.js
+++ b/platform/src/ActivityManager.js
@@ -354,12 +354,8 @@ class ActivityManager {
     interpolate(someString) {
         let result = someString;
 
-        // console.log (`Asked to interpolate ${someString}.`);
-
         for (let i = 0; i < sessionStorage.length; i++) {
             let currentKey = sessionStorage.key(i);
-
-            // console.log(`Interpolating for key ${currentKey}.`);
 
             if (currentKey !== "isAuthenticated") {
                 // TODO: This *assumes* this can only be a panel ID, but that may change over time,
@@ -368,7 +364,6 @@ class ActivityManager {
             }
         }
 
-        // console.log(`Result of interpolation: ${result}.`);
         return result;
     }
 
--- a/platform/src/ActivityManager.js
+++ b/platform/src/ActivityManager.js
@@ -360,7 +360,7 @@ class ActivityManager {
             if (currentKey !== "isAuthenticated") {
                 // TODO: This *assumes* this can only be a panel ID, but that may change over time,
                 // so this code may need to be improved to only allow access to panel IDs explicitly
-                result = result.replace("${ID-" + currentKey + "}", sessionStorage.getItem(currentKey));
+                result = result.replace("{{ID-" + currentKey + "}}", sessionStorage.getItem(currentKey));
             }
         }
 
--- a/platform/src/ActivityManager.js
+++ b/platform/src/ActivityManager.js
@@ -346,15 +346,40 @@ class ActivityManager {
         return null;
     }
 
+    /**
+     * Interpolate someString using information from session storage.
+     * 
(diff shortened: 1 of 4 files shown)


szschaler:
Documentation on the Wiki still needs updating when this PR is merged.

platform/src/ActivityManager.js
```diff
@@ -346,15 +346,40 @@ class ActivityManager {
...
 
+    /**
+     * Interpolate someString using information from session storage.
+     * 
+     * @param {*} someString the string to be changed
+     */
+    interpolate(someString) {
+        let result = someString;
+
+        // console.log (`Asked to interpolate ${someString}.`);
```
barnettwilliam, line 357: The commented out code for console logging should be removed. 

A single summary could be logged when a replacement is made so the values and their replacements can be seen.

szschaler: Agreed. I have removed the commented-out code. I haven't put in an active log statement as it seems a lot of logging for very little use...

platform/src/ActivityManager.js
```diff
@@ -346,15 +346,35 @@ class ActivityManager {
...
+    interpolate(someString) {
+        let result = someString;
+
+        for (let i = 0; i < sessionStorage.length; i++) {
+            let currentKey = sessionStorage.key(i);
 
+            if (currentKey !== "isAuthenticated") {
+                // TODO: This *assumes* this can only be a panel ID, but that may change over time,
+                // so this code may need to be improved to only allow access to panel IDs explicitly
+                result = result.replace("${ID-" + currentKey + "}", sessionStorage.getItem(currentKey));
```
barnettwilliam, line 363: Using `${...}` sytanx overlaps with bash and would prevent envsubst being used on activity files consider using an alternative e.g. `#{...}`. We don't currently use envsubst in this way.

barnettwilliam (commented):
Generally looks good, just the question on syntax overlap with bash.

platform/src/ActivityManager.js
szschaler: Good point. I've changed this to `{{ ... }}` in line with standard templating engines

barnettwilliam: LGTM

barnettwilliam (approved):
LGTM

PR #157, opened 2024-01-09
Be a bit more defensive about URL-loaded files

szschaler:
Checking that files loaded from a URL aren't null before accessing their contents seems a good idea :-)


barnettwilliam (approved):
LGTM

PR #161, opened 2024-01-10
Allow URL interpolation in tool URLs

szschaler:
This means that tools can be generated on the fly if needed, for example for Xtext.


barnettwilliam (approved):
LGTM

PR #162, opened 2024-01-12
Add support for URL rewriting in tool specifications

szschaler:
This PR adds support for tool specifications to use `{{BASE-URL}}` to refer to the URL from which the tool is deployed (strictly, the URL where the tool specification is hosted, minus the file name of the tool specification and any trailing `/`). This will automatically be rewritten when the tool specification is loaded, so that tool specifications can be written independently of location of deployment.


barnettwilliam (approved):
LGTM. Testing this higlighted an issue with the Xtext tool service endpoint structure but raised as - https://github.com/mdenet/platformtools/issues/37

PR #165, opened 2024-01-22
Provide a notification on editor builds failing 

barnettwilliam:
Provide a notification on editor builds failing for https://github.com/mdenet/platformtools/issues/39#issuecomment-1900066163

### DIFF
--- a/platform/src/Playground.js
+++ b/platform/src/Playground.js
@@ -1033,6 +1033,8 @@ async function checkEditorReady(statusUrl, editorInstanceUrl, editorPanelId, edi
         if (result.error){
             // Unsuccessful
             console.log("Editor failed start.");
+            Metro.notify.killAll();
+            notification("Build Failed", "An error occurred while building the editor.");
 
         } else if (!result.editorReady){
             await new Promise(resolve => setTimeout(resolve, 2000));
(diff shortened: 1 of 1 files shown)


platform/src/Playground.js
```diff
@@ -1033,6 +1033,8 @@ async function checkEditorReady(statusUrl, editorInstanceUrl, editorPanelId, edi
         if (result.error){
             // Unsuccessful
             console.log("Editor failed start.");
+            Metro.notify.killAll();
+            notification("Build Failed", "An error occurred while building the editor.");
```
szschaler, line 1037: This should be an `errorNotification` and it should also include `result.error` in the message.

This also needs to clear the `sessionStorage` element and to call `activityManager.setActivityVisibility` to make sure that the menu item disappears.

szschaler (changes requested):
See comment

platform/src/Playground.js
barnettwilliam: The convention that's being used  currently is that error notifications are for platform errors. 

Commit 8f9cd517f4f5197b8ee5790a40f563457e486394  displays amber coloured notifications for editor creation failing and the message received from the tool.  It also clears the storage and sets the visibility of the previously generated editor.

szschaler (approved):

PR #167, opened 2024-01-26
Added button for collecting user feedback.

barnettwilliam:
Added button for collecting user feedback, when deployed FEEDBAK_SURVEY_URL will be updated to point to the Microsoft forms survey [here](https://forms.office.com/Pages/ResponsePage.aspx?id=FM9wg_MWFky4PHJAcWVDVraRChPUQ-ZGrubhv7wZFeNUM0M4VVQ4SU9MRFVWUUFJSldWU05CUUNCWC4u). 

image

### DIFF
--- a/platform/public/css/index.css
+++ b/platform/public/css/index.css
@@ -22,3 +22,19 @@
     background-color: darkslategrey;
     color: aliceblue;
 }
+
+#feedback-button {
+    position: absolute !important;
+    z-index: 10000 !important;
+    padding: 12px;
+    border-radius: 20px;
+    box-shadow: 0 4px 8px 0 gray, 0 6px 20px 0 gray;
+    background-color: lightblue;
+    bottom: 5%;
+    right: 2%;
+    transition-duration: 0.4s;
+  }
+
+  #feedback-button:hover { 
+    background-color: #3498db;
+  }
--- a/platform/public/css/index.css
+++ b/platform/public/css/index.css
@@ -31,7 +31,7 @@
     box-shadow: 0 4px 8px 0 gray, 0 6px 20px 0 gray;
     background-color: lightblue;
     bottom: 5%;
-    right: 2%;
+    left: 2%;
     transition-duration: 0.4s;
   }
 
--- a/platform/public/index.html
+++ b/platform/public/index.html
@@ -16,8 +16,9 @@
 <body class="h-100" onresize="fit()" onload="fit();updateGutterVisibility();">
     <div id="preloader"><img src="images/preloader.gif" width="100px"></div>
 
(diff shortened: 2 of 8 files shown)


platform/public/css/index.css
```diff
@@ -21,4 +21,20 @@
...
+
+#feedback-button {
+    position: absolute !important;
+    z-index: 10000 !important;
+    padding: 12px;
+    border-radius: 20px;
+    box-shadow: 0 4px 8px 0 gray, 0 6px 20px 0 gray;
+    background-color: lightblue;
+    bottom: 5%;
+    right: 2%;
```
szschaler, line 34: This places this on top of content in the learning activity, which isn't ideal. Could we place it in the bottom-**left** corner instead?

platform/public/index.html
```diff
@@ -16,6 +16,12 @@
 <body class="h-100" onresize="fit()" onload="fit();updateGutterVisibility();">
     <div id="preloader"><img src="images/preloader.gif" width="100px"></div>
 
+    <div id='feedback-button'>
+        <p><a href="FEEDBAK_SURVEY_URL" target="_blank" rel="noopener noreferrer">
```
szschaler, line 20: `FEEDBAK` -> `FEEDBACK` 😄 

More importantly, though: how does this get instantiated effectively? I would like to ensure we don't create more and more places requiring manual adjustment on deployment. Can we automate filling this out on deployment, e.g. via a `.env` file?

szschaler (changes requested):
Generally looks good. 

In the MS Form: you have a nice triaging question at the start. I was expecting this to then lead to different sections so that you can ask specific follow-up questions depending on the type of feedback the student wants to provide.

platform/public/index.html
barnettwilliam: Corrected in f2298d7605d464f87af539998b569d29fd45ad73 and now settable via en environment variable. The feedback button is now only displayed if the env variable is set.

barnettwilliam:
> Generally looks good.
> 
> In the MS Form: you have a nice triaging question at the start. I was expecting this to then lead to different sections so that you can ask specific follow-up questions depending on the type of feedback the student wants to provide.

I have updated the form as suggested to be more selective based on the initial question. For the major issues I added an option to provide the activity for further investigation. I wanted to try to minimise the number of questions asked let me know if there is anything you think specifically would be useful to add, potentially some question on the system they were using e.g. browser, os etc.

szschaler (approved):
Looks good to me.

In the form, I would suggest keeping the "URL to the platform" link for all types of questions, but make it optional except where issues occurred.

barnettwilliam:
Great form updated

PR #169, opened 2024-01-30
Fix UTF-8 decoding errors

szschaler:
This PR fixes some issues with encoding / decoding of UTF-8, especially in relation to Base64 (see also https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem for explanation of the problem). This is important, for example, because Xtend code uses guillemets which are only available in UTF-8, not plain ASCII.

### DIFF
--- a/platform/src/FileHandler.js
+++ b/platform/src/FileHandler.js
@@ -10,6 +10,11 @@ class FileHandler {
     }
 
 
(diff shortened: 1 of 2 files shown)


barnettwilliam (approved):
LGTM

PR #170, opened 2024-01-31
Fix utf8 encoding on save

barnettwilliam:
Use utf8 encoding when saving panel contents to support non-ascii characters.


szschaler (approved):
LGTM

PR #175, opened 2024-02-09
Report all errors on playground start up

### LINKED ISSUE #128: If activity file cannot be found, no error message is shown to user
If an activity file cannot be found, [Line 342 of ActivityManager.js](https://github.com/mdenet/educationplatform/blob/242e76d439eedbef2b7e00543fc9656ab12bdf91/platform/src/ActivityManager.js#L342) will result in a `null` file and an exception will occur on the next line. This is invisible to the platform user.

Instead, an error message should be provided to the user.

szschaler:
This PR adds code to report all errors on playground startup.

Previous code only reported configuration errors and platform errors, using a specific format, but did not provide a catch-all section reporting any other errors that might occur. This now adds this catch-all block, which will report, for example, parsing errors and errors opening the activitiy file (which closes #128).

### DIFF
--- a/platform/src/Playground.js
+++ b/platform/src/Playground.js
@@ -245,6 +245,21 @@ function displayErrors(errors){
                 contentPanelDiv.append(content);
             });
         }
+
+        const otherErrors = errors.filter((e) => !(configErrors.includes(e) || platformErrors.includes(e)))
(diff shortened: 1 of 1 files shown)


barnettwilliam (approved):
LGTM

PR #176, opened 2024-02-09
Resolve activity buttons without tool buttons

szschaler:
This PR fixes a problem where activity-defined buttons would only be shown if there were any buttons defined in the panel definition, too.


barnettwilliam (approved):
LGTM

PR #178, opened 2024-02-13
Correctly report activity access issues

szschaler:
This PR fixes a problem whereby accessing a non-existing activities.json or a file where the permissions were missing would not lead to an error being reported to the user but would quietly crash the token server and the platform.

In the PR, the token server correctly reports the issue and the platform correctly raises an error to the user rather than trying to process the activity specification.

### DIFF
--- a/tokenserver/src/middleware/ErrorHandlingMiddleware.js
+++ b/tokenserver/src/middleware/ErrorHandlingMiddleware.js
@@ -2,8 +2,8 @@
 
 const errorHandlingMiddleware = (err, req, res, next) => {
   if (err.status != null) {
-    res.status(err.statusCode).json({
-        msg: err.message,
+    res.status(err.status).json({
+        msg: err.response.data.message,
         success: false,
       });
   } else {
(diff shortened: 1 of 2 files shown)


tokenserver/src/middleware/ErrorHandlingMiddleware.js
```diff
@@ -2,8 +2,8 @@
 
 const errorHandlingMiddleware = (err, req, res, next) => {
   if (err.status != null) {
-    res.status(err.statusCode).json({
-        msg: err.message,
+    res.status(err.status).json({
+        msg: err.response.data.message,
```
barnettwilliam, line 6: Not yet used?

barnettwilliam (commented):
Generally good one question

tokenserver/src/middleware/ErrorHandlingMiddleware.js
szschaler: I don't understand the question? This is where the error message is in an official response from GitHub...

szschaler: In fairness, it makes little difference atm other than for debugging purposes, as the platform file handler immediately ignores the message.

szschaler: So we could in principle simplify to `"File not found"` or something similar

PR #179, opened 2024-02-13
Correctly use ref when loading files via token server

szschaler:
This PR fixes the token server so that it correctly uses the provided `ref` parameter when loading files from GitHub.


szschaler:
Saving files did already use the `branch` parameter, so should continue to work correctly.

szschaler:
Depends on #178 which should be merged first.

barnettwilliam (approved):
LGTM

PR #185, opened 2024-02-23
Unit test code coverage

barnettwilliam:
Adds code coverage and reporting of unit tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,3 +57,22 @@ jobs:
           path: platform/reports/TESTS*.xml
           reporter: java-junit
           fail-on-error: true
+
+      - name: Setup .NET Core - for coverage report # Required to execute ReportGenerator
+        uses: actions/setup-dotnet@v3
+        if: success() || failure() && steps.run-unit-tests.outcome == 'success'
+        with:
+            dotnet-version: 8.x
+            dotnet-quality: 'ga'
+
+      - name: Create coverage md report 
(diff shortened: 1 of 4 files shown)


barnettwilliam:
@szschaler as demo'd in this mornings meeting code coverage is reported in the summary github ci worflow e.g. https://github.com/mdenet/educationplatform/actions/runs/8013383834

There wasn't a huge selection of GiHub actions for reporting code coverage nicely that worked, so went with adding markdown to the workflow action summary using action - https://github.com/danielpalme/ReportGenerator. 

Do we want to improve this report e.g. publishing a detailed html report with history as a webpage or do you think the simple output is sufficient?

Another option that offers the most comprehensive and visually appealing report is https://about.codecov.io/ that provide a GitHub action, however, it is an external service so it sends the results to them. It is free for opensource projects, not sure if that applies to us.

szschaler:
I think that simple output is enough for now.

I notice the following annotation:

```
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/setup-node@v3, actions/setup-dotnet@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

Is this something we can fix?

barnettwilliam:
Yes updating the `actions/checkout`, `actions/setup-dotnet` in my test repo fixes this so I'll update here.

szschaler (approved):
Looking good.

PR #186, opened 2024-02-29
Removed unused Backend class

barnettwilliam:
The Backend class from the Epsilon playground has been replaced with tool configurations files so it's no longer required.


szschaler (approved):
LGTM

PR #187, opened 2024-03-05
Unit tests for ActivityManager

barnettwilliam:
Some unit tests for ActivityManager.

A couple of minor changes were required for test-ability:

- Calls to global browser objects `document` and `window` should be via utility functions that can be stubbed.
- Imports have to be via objects and not direct export.  This is because ES6 protects exports which prevents them being spied on by Jasmine. It is a problem for all javascript test frameworks.
- ActivityManager constructor made costly http requests so these were separated out into initializeActivities() method that's called after it's created.

Code coverage for the module at approx 52%

### DIFF
--- a/platform/test/resources/TestUtility.js
+++ b/platform/test/resources/TestUtility.js
@@ -11,12 +11,3 @@ export function configObjectEquals(configA, configB){
 
     return (strConfigA === strConfigB)
 }
-
-/**
- * Prints a configuration object including its values.
- * @param {Object} config 
- * @returns {string} the string representation
- */
-export function configToString (config) {
-    return JSON.stringify(config);
-}
--- a/platform/test/resources/TestUtility.js
+++ b/platform/test/resources/TestUtility.js
@@ -1,0 +1,22 @@
+/**
+ * Compares two configuration objects. 
+ * @param {Object} configA
+ * @param {Object} configB
+ * @returns {Boolean} true if the objects values are equal.
+ */
+export function configObjectEquals(configA, configB){
+
+    const strConfigA = JSON.stringify(configA);
+    const strConfigB = JSON.stringify(configB);
+
+    return (strConfigA === strConfigB)
+}
+
+/**
+ * Prints a configuration object including its values.
+ * @param {Object} config 
+ * @returns {string} the string representation
+ */
+export function configToString (config) {
+    return JSON.stringify(config);
+}
--- a/platform/test/spec/testActivityManagerSpec.js
+++ b/platform/test/spec/testActivityManagerSpec.js
@@ -1,0 +1,133 @@
+/*global describe, it, expect, spyOn --  functions provided by Jasmine */
+import {ActivityManager} from "../../src/ActivityManager.js";
+
+import { FileHandler } from "../../src/FileHandler.js";
+import { ActivityConfigValidator } from "../../src/ActivityConfigValidator.js";
+import { utility } from "../../src/Utility.js";
+
+
+
+describe("ActivityManager", () => {
+
+        it("can be created", () => {
+            // Setup
+            const fileh = new FileHandler("test://th.url") 
+            const refPanelDef = () => {};
+                
+            // Call the target object
+            var am = new ActivityManager(refPanelDef , fileh);
+ 
+            // Check the expected results
+            expect(am).toBeInstanceOf(ActivityManager);
+        })
+ 
+        it("constructor -  initialises the configValidator property", () => {
+            // Setup
+            const fileh = new FileHandler("test://th.url") 
+            const refPanelDef = () => {};
+
+            // Call the target object
+            var am = new ActivityManager(refPanelDef , fileh);
+
+            // Check the expected results
+            expect(am.configValidator).toBeInstanceOf(ActivityConfigValidator);
+        })        
+
+        it("constructor -  initialises the accessPanelDef property using param panelDefAccessor", () => {
+            // Setup
+            const fileh = new FileHandler("test://th.url") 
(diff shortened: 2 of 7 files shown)


platform/test/resources/TestUtility.js
```diff
@@ -0,0 +1,22 @@
...
+export function configObjectEquals(configA, configB){
+
+    const strConfigA = JSON.stringify(configA);
+    const strConfigB = JSON.stringify(configB);
+
+    return (strConfigA === strConfigB)
+}
+
+/**
+ * Prints a configuration object including its values.
```
szschaler, line 16: The function doesn't strictly "print" the object, it just returns its stringification. Given that that's the case, is the function actually needed or could it just be replaced with direct uses of `JSON.stringify`?

platform/test/spec/testActivityManagerSpec.js
```diff
@@ -0,0 +1,264 @@
...
+import { ActivityConfigValidator } from "../../src/ActivityConfigValidator.js";
+import { utility } from "../../src/Utility.js";
+import { ACTIVITY_2PANELS_1ACTION } from "../resources/TestActivityFiles.js";
+import { TOOL_1PANELDEF_1FUNCTION } from "../resources/TestToolFiles.js";
+import { configObjectEquals, configToString } from "../resources/TestUtility.js";
+
+
+describe("ActivityManager", () => {
+
+    it("can be created", () => {
```
szschaler, line 14: Given that there are multiple tests for the constructor, does it make sense to group them into a nested `describe` and possibly even reuse the constructed object (with a `beforeAll`)?

szschaler (changes requested):
Looks good. Just some minor suggestions.

platform/test/resources/TestUtility.js
barnettwilliam: Yes it's not particularly useful as it is so will use stringify directly.

barnettwilliam: Removed in 52769f3d17c78207d5778a9ec46d8e090149bfb0

platform/test/spec/testActivityManagerSpec.js
barnettwilliam: Fixed in 52769f3d17c78207d5778a9ec46d8e090149bfb0

platform/test/spec/testActivityManagerSpec.js
```diff
@@ -6,91 +6,67 @@ import { ActivityConfigValidator } from "../../src/ActivityConfigValidator.js";
...
-    it("constructor -  initialises the accessPanelDef property using param panelDefAccessor", () => {
-        // Setup
-        const fileh = new FileHandler("test://th.url") 
-        const refPanelDef = () => {};
+            spySearch = spyOn( utility, "getWindowLocationSearch");
 
-        // Call the target object
-        var am = new ActivityManager(refPanelDef , fileh);
+            // Call the target object
+            am = new ActivityManager(refPanelDef, fileh);
```
szschaler, line 30: Should this be in each test (there is at least one test that has to call this again anyway...)?

platform/test/spec/testActivityManagerSpec.js
```diff
@@ -6,91 +6,67 @@ import { ActivityConfigValidator } from "../../src/ActivityConfigValidator.js";
...
+        })
 
-    it ("constructor -  if the current activity is provided in the url query string, set the activityId property", () => {
-        // Setup
-        const ACTIVITY_URL="test://a.url";
-        const ACTIVITY_ID="a1"
-        const QUERY = "?"+ ACTIVITY_ID + "=&" + "activities=" + ACTIVITY_URL;
-        const fileh = new FileHandler("test://th.url") 
-        const refPanelDef = () => {};
+        it ("if the current activity is provided in the url query string, set the activityId property", () => {
```
szschaler, line 60:
```suggestion
        it ("sets the activityId property if the current activity is provided in the url query string", () => {
```

platform/test/spec/testActivityManagerSpec.js
```diff
@@ -6,91 +6,67 @@ import { ActivityConfigValidator } from "../../src/ActivityConfigValidator.js";
...
-        // Call the target object
-        var am = new ActivityManager(refPanelDef , fileh);
+            expect(am.activityId).toBe(ACTIVITY_ID);
+        })
 
-        // Check the expected results
-        expect(am.activityId).toBe(ACTIVITY_ID);
     })
 
     it("initializeActivities - causes the activity file to be fetched from its URL using the fileHandler", () => {
```
szschaler, line 72: Shouldn't these be in the next `describe` as they are about activity initialisation?

szschaler (changes requested):
Thanks, much improved. Some other questions/suggestions...

platform/test/spec/testActivityManagerSpec.js
barnettwilliam: Yes it does make each test uniform fixed in 023439982f88091a37a89567f047fcb5495458bd

platform/test/spec/testActivityManagerSpec.js
barnettwilliam: Yes the test style evolved slightly as they were added it makes sense to include them in the same describe block. Fixed in 023439982f88091a37a89567f047fcb5495458bd

platform/test/spec/testActivityManagerSpec.js
barnettwilliam: Updated in c1fc48ec710862c2619fcad8fc297215a00334e8

szschaler (approved):
LGTM

Issue #192, opened 2024-03-12
ToolsManager getActionFunction deprecated


barnettwilliam:
The getActionFunction should moved to the FunctionRegistry class. This requires some rework of the FunctionRegistery to use ActionFunction and potentially adding some helper methods to ActionFunction e.g getInputParameterTypes  rather than pass them in to FunctionRegistry's  constructor. 

Related to #40.

Issue #40, opened 2023-04-25
Refactor function calling code by moving it into ToolManager or FunctionRegistry


szschaler:
This is about moving the code for calling action functions (and probably for conversion of parameters) into `ToolManager` or even into the `FunctionRegistry`. 

This depends on the refactoring in #34 being completed first, so that there actually _is_ a function to move.

PR #193, opened 2024-03-12
Refactor tool function calling code

barnettwilliam:
Long awaited changes for #40 refactor tool function calling code. Mostly moving tool function and conversion function invocation to `ToolsManager` and `FunctionRegistry`.

There is still moving the `getActionFunction()` from `ToolsManager` into `FunctionRegistry` todo  but raised as issue #192 because it requires some rework of `FunctionRegistry`


szschaler (changes requested):
I would feel better about this refactoring if we first had a robust set of tests around calling action functions and conversions using the current implementation before we move the code (and its corresponding, but not established) tests,

barnettwilliam:
Ok, I can write tests for the existing implementation and then use them here to verify these changes

platform/src/EducationPlatformApp.js
```diff
@@ -895,7 +620,7 @@ class EducationPlatformApp {
             //actionRequestData.outputLanguage = outputLanguage;
 
         // Call backend conversion and service functions
-        let actionResultPromise = this.invokeActionFunction(buttonConfig.actionfunction, parameterMap);
+        let actionResultPromise = this.toolsManager.invokeActionFunction(buttonConfig.actionfunction, parameterMap);
```
szschaler, line 623: Not here, but slightly above: we call the `toolsManager` to get the actual function object. We only use this to determine the type of the `language` parameter. Is there way to refactor the code so that we don't have to access the function object here at all?

platform/src/ToolsManager.js
```diff
@@ -1,21 +1,25 @@
...
     tools = {};
     functionRegister;
+    errorNotification;
 
-    constructor(){
+    /**
+     * @param {function(String)} errorNotifier - the function to call to display an error
+     */
+    constructor(errorNotifier){
         this.configValidator= new ToolConfigValidator();
```
szschaler, line 20:
```suggestion
        this.configValidator = new ToolConfigValidator();
```

platform/src/ToolsManager.js
```diff
@@ -1,21 +1,25 @@
...
-        this.functionRegister= new FunctionRegistry();
+        this.functionRegister= new FunctionRegistry(this);
```
szschaler, line 21:
```suggestion
        this.functionRegister = new FunctionRegistry(this);
```

platform/test/spec/testFunctionRegistrySpec.js
```diff
@@ -1,15 +1,19 @@
+/*global describe, it, expect, beforeEach, afterEach, expectAsync  --  functions provided by Jasmine */
+/*global jasmine --  object provided by Jasmine */
+
 import {FunctionRegistry} from "../../src/FunctionRegistry.js"
+import { ActionFunction } from "../../src/ActionFunction.js";
 
 describe("FunctionRegistry", () => {
 
-    it("a registered function can be looked up", () => {
+    it("a registered function can be found", () => {
```
szschaler, line 9:
```suggestion
    it("can find a registered function", () => {
```

szschaler (approved):
Some small comments, but overall LGTM

platform/src/EducationPlatformApp.js
barnettwilliam: Since the language type is always expected to be 'text' it could be set directly adding a constant `ACTION_FUNCTION_LANGUAGE_TYPE = "text"` which saves having to access it via `toolActionFunction`.

Quickly trying this out has also revealed that the OCL tool sets the language type to string so this would need to be update to use text to.

szschaler: Sounds like a sensible thing to do.

barnettwilliam: Fixed in bcf669c333ac5b4bca5da6db8d202f322718398f and also requires change in PR thttps://github.com/mdenet/platformtools/pull/66 to be merged.

szschaler (approved):

PR #194, opened 2024-03-13
Class based platform for unit testing

barnettwilliam:
Moved Playground.js into an `EducationPlatform` class to facilitate unit testing. Platform setup is now done by index.js that instantiates the `EducationPlatform`.  This change makes it possible to unit test the methods used by platform without the platform setup conflicting with the tests.

Tested and working using recent activities with composite panels including: Xtext editor, loading from a private repo, saving to a repo. Tested and working using Linux Firefox v123 and Chrome v122 and Windows Edge v122.


platform/src/index.js
```diff
@@ -0,0 +1,23 @@
...
+window.fit = platform.fit.bind(platform);
+window.updateGutterVisibility = platform.updateGutterVisibility.bind(platform);
+window.runAction = platform.runAction.bind(platform);
+window.panels = platform.panels;
+window.savePanelContents = platform.savePanelContents.bind(platform);
+window.toggle = platform.toggle.bind(platform);
+window.togglePanelById = platform.togglePanelById.bind(platform);
+//window.renderDiagram = renderDiagram;
+window.longNotification = platform.longNotification.bind(platform);
+window.getPanelTitle = platform.getPanelTitle.bind(platform);
```
szschaler, line 23: I guess it would be cleaner to rewrite those onClick handlers to access the platform object and interact with it directly...

platform/test/spec/testToolManagerSpec.js
```diff
@@ -44,7 +44,7 @@ describe("ToolManager", () => {
             // Check the expected results
             expect(tm.toolsUrls).toHaveSize(TOOL_URLS.length);
             
-            for (const i in TOOL_URLS){
+            for (let i = 0; i < TOOL_URLS.length; i++ ){
```
szschaler, line 47: Why is this change needed?

szschaler (approved):
Some thoughts and suggestions added, but ok overall.

platform/src/index.js
barnettwilliam: I think it would be better to not use window globals by registering these in the platform class but this should be a separate change.

I would prefer if the platform object were not added to the windows globals as there is the potential for it to clash and be modified inadvertently by other libraries. It would be good to namepace these current ones to something like `mdenetEducationPlatform`

platform/src/EducationPlatform.js
```diff
@@ -0,0 +1,1069 @@
+/*global $ -- jquery is exterally imported*/
+/*global FEEDBACK_SURVEY_URL -- is set by environment variable*/
+/*global Metro -- Metro is externally imported*/
+
+import * as ace from 'ace-builds/src-min-noconflict/ace';
+import 'ace-builds/src-min-noconflict/theme-eclipse';
+import 'ace-builds/src-min-noconflict/mode-xml';
+import 'ace-builds/src-min-noconflict/mode-yaml';
+import 'ace-builds/src-min-noconflict/mode-java';
+import 'ace-builds/src-min-noconflict/mode-html';
```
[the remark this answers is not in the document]
barnettwilliam: They are required for code highlighting in the OutputPanel e.g. for generated code

platform/test/spec/testToolManagerSpec.js
barnettwilliam: The `for ... in` meant things other than array elements were being checked e.g. functions from the Array class using index only the elements are checked. I couldn't determine why after importing the playground there was difference in array objects created by the test compared to the one returned by the platform.

platform/src/index.js
szschaler: I see. Would be good to add an issue to clean this up: really there should be a way for action definitions to specify the languages to be used in output panels and that should be respected -- these output languages and their highlighting rules should then perhaps be loaded from a tool definition, though some might be provided centrally via the common tools specification.

platform/test/spec/testToolManagerSpec.js
szschaler: Understood. This is one of the peculiarities of JS that I will never really get used to. Good to have this cleaned up.

platform/src/EducationPlatform.js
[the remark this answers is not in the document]
szschaler: I see. Would be good to add an issue to clean this up: really there should be a way for action definitions to specify the languages to be used in output panels and that should be respected -- these output languages and their highlighting rules should then perhaps be loaded from a tool definition, though some might be provided centrally via the common tools specification.

platform/src/index.js
szschaler: That sounds like an even better solution. I agree that we shouldn't fix this here, but we should have an issue that allows us to track the problem and prioritise it separately.

platform/src/EducationPlatform.js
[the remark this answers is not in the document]
barnettwilliam: Agree, raised issue https://github.com/mdenet/educationplatform/issues/195

platform/src/index.js
barnettwilliam: Raised issue https://github.com/mdenet/educationplatform/issues/196 to capture this.

szschaler (approved):

PR #200, opened 2024-03-20
Fix only provide parameters used by tool functions

### LINKED ISSUE #41: Clean up Epsilon Wrapper so that functions only expect the parameters they are actually using

barnettwilliam:
Modified invokeActionFunction() to not populate unused parameters as 'undefined' because this is no longer necessary when using the Epsilon micronaut tool service.

Along https://github.com/epsilonlabs/playground-micronaut/pull/4 with fixes #41 

Mentioned in code review for PR https://github.com/mdenet/educationplatform/pull/198#discussion_r1531961177

### DIFF
--- a/platform/src/EducationPlatformApp.js
+++ b/platform/src/EducationPlatformApp.js
@@ -491,11 +491,7 @@ class EducationPlatformApp {
 
                     const panelConfig = parameterMap.get(param.name); 
 
-                    if (panelConfig == undefined){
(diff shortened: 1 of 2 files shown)


szschaler (approved):
LGTM: do we need to wait for your PR in the Epsilon micronaut before merging this?

barnettwilliam:
There's no need to wait for the Epsilon micronaut PR

PR #203, opened 2024-04-08
Error handling

barnettwilliam:
Common handling of errors and exceptions via ErrorHandler.

Unexpected error notifcation are displayed for any unhandled exceptions via `window.onerror()`  that ErrorHandler initialises to call `notify()`.

### DIFF
--- a/platform/src/EducationPlatformApp.js
+++ b/platform/src/EducationPlatformApp.js
@@ -557,10 +557,10 @@ class EducationPlatformApp {
             
             if(action.source.buttons){
                 //Buttons defined by activity
-                buttonConfig=  action.source.buttons.find( btn => btn.id == sourceButton );
+                buttonConfig = action.source.buttons.find (btn => btn.id == sourceButton);
             } else {
                 //Buttons defined by tool
-                buttonConfig= action.source.ref.buttons.find( btn => btn.id == sourceButton );
+                buttonConfig = action.source.ref.buttons.find (btn => btn.id == sourceButton);
             }  
 
             // Create map containing panel values
--- a/platform/src/EducationPlatformApp.js
+++ b/platform/src/EducationPlatformApp.js
@@ -550,55 +550,63 @@ class EducationPlatformApp {
 
         // Get the action
         var action = this.activityManager.getActionForCurrentActivity(source, sourceButton);
-        
-        let buttonConfig;
-        if(action.source.buttons){
-            //Buttons defined by activity
-            buttonConfig=  action.source.buttons.find( btn => btn.id == sourceButton );
+       
+        if (!action){
+            let err = new EducationPlatformError(`Cannot find action given panel '${source}' and button '${sourceButton}'`);
+            this.errorHandler.notify("Failed to invoke action.", err);
+
         } else {
-            //Buttons defined by tool
-            buttonConfig= action.source.ref.buttons.find( btn => btn.id == sourceButton );
-        }  
-
-        // Create map containing panel values
-        let parameterMap = new Map();
-
-        for (let paramName of Object.keys(action.parameters)){
-
-            let param = {};
-            const panelId = action.parameters[paramName].id;
+            // Action found so try and invoke
+            let buttonConfig;
             
-            if (panelId) { 
-                const panel = this.activityManager.findPanel(panelId, this.panels);
-                param.type = panel.getType();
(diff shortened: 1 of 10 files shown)


platform/src/EducationPlatformApp.js
```diff
@@ -152,7 +155,7 @@ class EducationPlatformApp {
 
         if (errors.length==0){
             // An activity configuration has been provided
-            this.toolsManager = new ToolsManager(this.errorNotification);
+            this.toolsManager = new ToolsManager(this.errorHandler.notify.bind(this.errorHandler));
```
szschaler, line 158: This is quite complicated "magic". Is there a way of simplifying this?

platform/src/EducationPlatformApp.js
```diff
@@ -576,59 +546,63 @@ class EducationPlatformApp {
...
         } else {
-            //Buttons defined by tool
-            buttonConfig= action.source.ref.buttons.find( btn => btn.id == sourceButton );
-        }  
+            // Action found so try and invoke
+            let buttonConfig;
+            
+            if(action.source.buttons){
+                //Buttons defined by activity
+                buttonConfig=  action.source.buttons.find( btn => btn.id == sourceButton );
```
szschaler, line 560:
```suggestion
                buttonConfig = action.source.buttons.find (btn => btn.id == sourceButton);
```

(I know this isn't part of this PR, but might as well clean it up while we're here.)

platform/src/EducationPlatformApp.js
```diff
@@ -576,59 +546,63 @@ class EducationPlatformApp {
...
+            } else {
+                //Buttons defined by tool
+                buttonConfig= action.source.ref.buttons.find( btn => btn.id == sourceButton );
```
szschaler, line 563:
```suggestion
                buttonConfig = action.source.ref.buttons.find (btn => btn.id == sourceButton);
```
ditto

szschaler (approved):
Generally looks good, just some small suggestions.

platform/src/EducationPlatformApp.js
barnettwilliam: @szschaler the `errorNotification()` etc could be moved to `PlaygroundUtility.js` and imported as required from there and means that it doesn't have to be passed into `ErrorHandler`

barnettwilliam: Simplified by moving notification functions to `PlaygroundUtility` in bbabad1ec6f1ea7a340b766d1c5ce449b0c08ed7. 

Rename `PlaygroundUtility` as a separate task

platform/test/spec/testEducationPlatformAppSpec.js
```diff
@@ -169,91 +169,4 @@ describe("EducationPlatformApp", () => {
             })
         })
     })
-
-
-    describe("notification()", () => {
```
szschaler, line 140: What's happened to all of these tests? Wouldn't it make sense to move them to the PlatformUtil tests spec?

platform/test/spec/testErrorHandlerSpec.js
```diff
@@ -1,27 +1,20 @@
...
+import { PlaygroundUtility } from "../../src/PlaygroundUtility.js";
 
 describe("ErrorHandler", () => {
 
     // Setup
     let notifierSpy;
 
     beforeEach(()=>{
-        notifierSpy = jasmine.createSpy("notifier");
+        notifierSpy = spyOn(PlaygroundUtility,"errorNotification");
```
szschaler, line 14: This is still being passed in as a parameter to the `ErrorHandler` constructor calls below. Doesn't that need fixing, too?

szschaler (changes requested):

platform/test/spec/testEducationPlatformAppSpec.js
barnettwilliam: Yes, these were moved but the file was missing from the commit.  Fixed in b4d4f4a293d734496827ac955323ea46d437b208

platform/test/spec/testErrorHandlerSpec.js
barnettwilliam: Thanks yes, fixed in 27027ebcef16eba9977cddc6e63726d01cdae3a0

szschaler (approved):

PR #205, opened 2024-04-11
Fix errors on resize

barnettwilliam:
Fixed creation of panel splitters in Layout so that panel layouts with an odd number of panels do not have an unnecessary splitter that causes a Metro exception on page resize, and added defensive check for fit() when no splitters are found.

### DIFF
--- a/platform/src/Layout.js
+++ b/platform/src/Layout.js
@@ -15,7 +15,7 @@ class Layout {
         var splitter;
 
         if ( panels.length == 1 ) {
-            splitter = Layout.createVerticalSplitter([panels[0].getElement()]);
+            splitter = panels[0].getElement(); // only the element to add
 
         } else if ( panels.length == 2  ) {
             splitter = Layout.createVerticalSplitter([panels[0].getElement(), panels[1].getElement()], "50, 50" );
@@ -36,7 +36,7 @@ class Layout {
 
             // Odd number of panels so add the last element
             if (panelsToLayout.length==1){
-                verticalSplitters.push( Layout.createVerticalSplitter([panelsToLayout.pop().getElement()]) );
+                verticalSplitters.push(panelsToLayout.pop().getElement());
             }
 
             splitter = Layout.createHorizontalSplitter( verticalSplitters, splitProportions);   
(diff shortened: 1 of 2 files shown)


platform/src/Layout.js
```diff
@@ -89,7 +89,13 @@ class Layout {
 
             // Create the splitters for the row
             let splitProportions = ("10, ".repeat(numberOfRows)); // Add a proportion per splitter
-            verticalSplitters.push ( Layout.createVerticalSplitter( panelsToLayout.map( pn => pn.getElement() ), splitProportions) );
+
+            if (panelsToLayout.length > 1) {
+                verticalSplitters.push( Layout.createVerticalSplitter( panelsToLayout.map( pn => pn.getElement() ), splitProportions) );
+            } else {
+                // No splitter required for a single panel
+                verticalSplitters.push( panelsToLayout[0].getElement() );
```
szschaler, line 97: Doesn't this just create another problem on Line 107? (BTW, why **do** we create another vertical splitter there?)

szschaler (changes requested):

platform/src/Layout.js
```diff
@@ -15,7 +15,7 @@ class Layout {
         var splitter;
 
         if ( panels.length == 1 ) {
-            splitter = Layout.createVerticalSplitter([panels[0].getElement()]);
+            splitter = panels[0].getElement(); // only the element to add
```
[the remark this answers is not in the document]
barnettwilliam: The naming of the dom page element "splitter" on line 45 is confusing and unrelated to the Metro element splitter, it is more accurately described as a container of panels that seems to only be used by `fit()` to resize panels on their size changing. Therefore, renaming the attribute that is applied on line 46 to something like `navview-content-panels` would be clearer without ant structural change to the page also some DRYing up as the same attributes are applied in `createFrom2dArray()`

platform/src/Layout.js
barnettwilliam: Good point, the splitter created on line 107 shouldn't be required I'll test with a single column example

platform/src/Layout.js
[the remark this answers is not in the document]
barnettwilliam: Improved in 0e92036a8e2da4919b76f716a95f5f193b2c8e14

platform/src/Layout.js
barnettwilliam: Fixed in 7ab53c7158eba08f7e2d5af13427ad60a902c85b

szschaler (approved):

PR #213, opened 2024-11-21
Base URL Fix

aryaei2000:
This is to remove the need of having to manually set base URLs on occasions where the platform needs to be accessed by an FQDN.


platform/src/ToolsManager.js
```diff
@@ -31,8 +31,26 @@ class ToolManager {
                 
                 let toolUrl = new Object();
                 toolUrl.id = ""; // Populate when tool is fetched
-                toolUrl.url = url;
-
+                if (this.isValidUrl(url)){
```
szschaler, line 34: This does two things at once, which need different treatment. We should have a function checking if there is a use of `{{BASE-URL}}` and a separate function checking validity of the URL. Suitable error checking should be used, though note that in certain cases (editor generation) invalid URLs are initially expected and shouldn't stop proceedings.

platform/src/ToolsManager.js
```diff
@@ -31,8 +31,26 @@ class ToolManager {
...
+                    let url_tail = url.split('/')[1];
+                    let url_port = this.getPort(url);
+                    if (url_port != null){
+                        let path = this.fetchPathByPort(url_port);
+
+                        if(path != null){
+                            let base_url = utility.getBaseURL();
+                            path = path.endsWith('/') ? path : path + '/';
+                            toolUrl.url = base_url + path + url_tail;
+                        }
```
szschaler, line 47: There should be an error notification here.

platform/src/ToolsManager.js
```diff
@@ -43,6 +61,50 @@ class ToolManager {
...
+        }
+
+        return null;
+    };
+
+    /**
+     * Communicates with the discovery service to fetch the relevant path by the service port.
+     * @returns string|null
+     */
+    fetchPathByPort(port) {
```
szschaler, line 84: Given that the discovery service only provides static data, why do we need to run it as a separate service at all? Couldn't the lookup simply happen in this function directly?

platform/src/ToolsManager.js
```diff
@@ -70,9 +132,25 @@ class ToolManager {
...
+                var toolConfig;
+                let response_text =  xhr.responseText;
+                let response_text_regexp = response_text.match(new RegExp(/{{BASE-URL}}:[0-9]*/));
+                if (response_text_regexp != null){
+                    // There is a reference to another endpoint, so base URL must be set accordingly (There is a {{BASE-URL}} placeholder plus a port).
+                    var base_url = utility.getBaseURL();
+                    let url_port = response_text_regexp[0];
+                    let path = this.fetchPathByPort(this.getPort(url_port));
+                    
+                    toolConfig = response_text.replaceAll(url_port, base_url + path);
```
szschaler, line 145: But what if there are multiple references to different endpoints, including, perhaps to this tool? I think you need to wrap your if statement into a loop and execute it separately for each match of the regexp.

szschaler (changes requested):
A few comments to look at. Why does the discovery service need to run as a separate service in the first place?

discovery-service/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+FROM node:19-bullseye AS dsbundle
+
+# Create mdenet-ep directory
+WORKDIR /usr/src/discovery-service
+
+# Install app dependencies
+# A wildcard is used to ensure both package.json AND package-lock.json are copied
+# where available (npm@5+)
+COPY package*.json ./
```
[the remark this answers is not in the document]
aryaei2000: Service discovery is now incorporated in education platform and runs as front-end code.

discovery-service/package.json
```diff
@@ -0,0 +1,15 @@
...
+    "name": "discovery-service",
+    "version": "1.0.0",
+    "description": "",
+    "main": "index.js",
+    "scripts": {
+      "test": "echo \"Error: no test specified\" && exit 1"
+    },
+    "keywords": [],
+    "author": "",
+    "license": "ISC",
```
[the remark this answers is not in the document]
aryaei2000: Service discovery is now incorporated in education platform and runs as front-end code.

platform/src/ToolsManager.js
aryaei2000: Service discovery is now incorporated in education platform and runs as front-end code.

platform/src/ToolsManager.js
aryaei2000: isValidUrl method is separated into two methods, `isUrlPlaceHolder()` checks whether the url is a placeholder or not.

platform/src/ToolsManager.js
aryaei2000: Added.

platform/src/ToolsManager.js
aryaei2000: Not sure I understood this correctly but this code block is already in a loop and on each iteration we only have one url variable. If you are referring to cases where we have {{ID-somthing}} as a placeholder, these are kept intact until the program reaches the `interpolate() `method in the `ActivityManager` class.

platform/src/ToolsManager.js
```diff
@@ -70,9 +132,25 @@ class ToolManager {
             }
 
             if (xhr.status === 200) {    
-                // Rewrite URLs in tool config
-                let baseURL = toolUrl.url.substring(0, toolUrl.url.lastIndexOf("/")); // remove the name of the json file (including the trailing slash)
-                let toolConfig = xhr.responseText.replaceAll("{{BASE-URL}}", baseURL);
+
+                var toolConfig;
+                let response_text =  xhr.responseText;
+                let response_text_regexp = response_text.match(new RegExp(/{{BASE-URL}}:[0-9]*/));
```
szschaler, line 161:
```suggestion
                let response_text_regexp = response_text.match(new RegExp(/{{BASE-URL}}(:[0-9]+)?/));
```
Then loop over all matches and for each one decide individually what the correct replacement is.

platform/src/ToolsManager.js
aryaei2000: Code block was refactored in a method and unit tests were added too.

platform/src/ToolsManager.js
aryaei2000: Regular expression was replaced as advised and now matches URL placeholders regardless of port presence.

aryaei2000:
The only remaining issue is an unnecessary escape character in the lint report which I have no idea on why it's being caused. The forward slashes should be escaped and if the backslashes are removed the regular expression doesn't function as expected.
[See here](https://github.com/mdenet/educationplatform/blob/eb97de3a6c3ca996e03c18ef9e5786e5979ea098/platform/src/ToolsManager.js#L127)

platform/src/ToolsManager.js
```diff
@@ -31,10 +31,8 @@ class ToolManager {
                 
                 let toolUrl = new Object();
                 toolUrl.id = ""; // Populate when tool is fetched
-                if (this.isValidUrl(url)){
-                    toolUrl.url = url;
-                }
-                else{
+                if (this.isUrlPlaceHolder(url)){
```
szschaler, line 34: But this also returns true for `{{ID-...}}` rewrite requests, so those would be treated as if they are `{{BASE-URL}}` occurrences. That seems wrong.

platform/src/ToolsManager.js
```diff
@@ -50,6 +48,14 @@ class ToolManager {
                         toolUrl.url = url;
                     }
                 }
+                else if (this.isValidUrl(url)){
```
szschaler, line 51: Shouldn't this case also capture `{{ID-...}}` cases?

platform/src/ToolsManager.js
```diff
@@ -178,6 +186,37 @@ class ToolManager {
...
+    /**
+     * Reads toocl config text and replaces {{BASE-URL}} placeholders with the tool URL
+     * @returns string
+     */
+    rewriteUrl(base_url, toolUrl, tool_config_string) {
+        let tool_config_regexp = tool_config_string.match(new RegExp(/{{BASE-URL}}(:[0-9]+)?/g));
+        var toolConfig = tool_config_string;
+
+        if (tool_config_regexp != null){
+            // There are references to another endpoint, so base URL must be set accordingly (There is a {{BASE-URL}} placeholder plus a port).
```
szschaler, line 198:
```suggestion
            // There are references to another endpoint, so base URL must be set accordingly (There is a {{BASE-URL}} placeholder plus an optional port).
```

szschaler: Actually, looking further down, it looks like you really want to make sure your regexp only matches `{{BASE-URL}}:<port>` because that then leaves only the "un-ported" `{{BASE-URL}}` cases.

platform/src/ToolsManager.js
```diff
@@ -178,6 +186,37 @@ class ToolManager {
...
+            for (const url_placeholder of base_url_placeholders.values()) {
+                let url_port = url_placeholder;
+                let port = this.getPort(url_port);
+                if (port != null){
+                    let path = this.fetchPathByPort(port);
+                    toolConfig = toolConfig.replaceAll(url_port, base_url + path);
+                }
+            }
+
+            // No port comes with the remaining placeholders, replace it with the tool reletive URL
```
szschaler, line 210:
```suggestion
            // No port comes with the remaining placeholders, replace it with the tool relative URL
```

platform/test/spec/testToolManagerSpec.js
```diff
@@ -60,6 +60,72 @@ describe("ToolManager", () => {
...
+
+        it("isUrlPlaceHolder - check if a string is a url placeholder", () => {
+            var url = "{{BASE-URL}}"
+            expect(tm.isUrlPlaceHolder(url)).toBe(true)
+
+            var url = "{{BASE-URL}}:123"
+            expect(tm.isUrlPlaceHolder(url)).toBe(true)
+
+            var url = "{{ID-panel-turtles}}"
+            expect(tm.isUrlPlaceHolder(url)).toBe(true)
```
szschaler, line 123: Should it be?

platform/test/spec/testToolManagerSpec.js
```diff
@@ -79,7 +145,15 @@ describe("ToolManager", () => {
             for (let key of EXPECTED_TOOL_KEYS){
                 expect( storedToolKeys.find(n => n===key ) ).toEqual(key);
             }
-        })
+        });
+
+        it("rewrites tool urls", () => {
+            var toolConfigWithPlaceholder = "somestring   {{BASE-URL}}somestring somestring {{BASE-URL}}:8070";
+            var toolConfigWithUrls = "somestring   http://toolurl somestring somestring http://toolurl:8080";
```
szschaler, line 152: Add an example with `{{ID-...}}` in it.

szschaler (changes requested):
A few questions remain

szschaler:
> The only remaining issue is an unnecessary escape character in the lint report which I have no idea on why it's being caused. The forward slashes should be escaped and if the backslashes are removed the regular expression doesn't function as expected. [See here](https://github.com/mdenet/educationplatform/blob/eb97de3a6c3ca996e03c18ef9e5786e5979ea098/platform/src/ToolsManager.js#L127)

The two that it flags are occurrences of `\/` inside a character class (i.e., between `[` and `]`), where presumably the escaping isn't needed because the regexp parser can work out it's not the end of the regexp. I note that other occurrences of `\/` in the same regexp are accepted by the linter.

platform/src/ToolsManager.js
aryaei2000: The reason to match the {{BASE-URL}}:port first is that any occurrences of these will be replaced with actual URLs before we get to occurrences of {{BASE-URL}} only. If done the other way round, `string.replaceAll()` function in JS will replace all instances of a placeholder and we cannot use the ports to determine the correct path.

szschaler: That's understood. The point I'm making is that your regexp shouldn't have a `?` after the port because you only want to match occurrences that have a port.

aryaei2000:
Linter is fixed thanks for the point!

platform/src/ToolsManager.js
szschaler: Yes, when `{{ID-***}}` placeholders are interpolated, then they won't get to this point at all, so that's not a problem.

**However**, when they aren't then they get here and there really isn't any valid substitution -- this is an expected situation not an error. Substituting as if it was a `{{BASE-URL}}` is definitely not valid. What should happen is that the case is recognised and ignored, with the containing activity being disabled. I think `EducationPlatform` manages that, in fact.

aryaei2000: We don't currently have a check at https://github.com/mdenet/educationplatform/blob/fe88277fd9cca3e4b9adbfcc4448e10423c828cd/platform/src/ToolsManager.js#L58

As you can see. If the `isValidURL()` method only returns true for actual URLs and not placeholders, then we can add a guard to skip `{{ID-...}}` at this stage. When they get interpolated, they will be a valid URL and can pass through.

aryaei2000: An error is pushed to the UI in `setToolsUrls()` when a candidate URL is neither a placeholder, nor a valid URL. Other cases it can throw an error is perhaps when doing string manipulation.

szschaler: Yes. The point I'm making is that unresolved `{{ID-***}}` tool URLs shouldn't lead to such an error being pushed, but should also not be rewritten as if they were `{{BASE-URL}}:port` URLs instead.

platform/test/spec/testToolManagerSpec.js
szschaler: BTW, should this also test if an error is pushed?

platform/src/ToolsManager.js
aryaei2000: As discussed `isUrlPlaceHolder() ` is refactored into `isBaseUrlPlaceHolder()` and `isIDPlaceHolder()` and logic around extracting URL tail is improved.

szschaler (approved):
LGTM

platform/test/spec/testToolManagerSpec.js
szschaler: This hasn't happened, has it?

platform/src/ToolsManager.js
aryaei2000: A new case is added to handle {{ID-...}} placeholders.

PR #217, opened 2025-01-23
212 performance issue

### LINKED ISSUE #212: Performance issue
I just tried to get somewhere around 40 students to generate an Xtext grammar at the same time and it crashed the server. As far as I can tell, the Xtext container and possibly even the Docker daemon got killed. The Xtext container would attempt to restart automatically, which would create extra load on the server, together with all clients continuing to poll for console updates.

It would be worth doing some performance testing to work out what we can do to improve this. It may be an issue to do with the polling for successful editor generation, which we should probably change to server-side events / websockets to improve efficiency.

aryaei2000:
Fixes #212

### DIFF
--- a/platform/src/EducationPlatformApp.js
+++ b/platform/src/EducationPlatformApp.js
@@ -463,8 +463,7 @@ class EducationPlatformApp {
                     // Language workbench
                     PlaygroundUtility.longNotification("Building editor");
 
-                    var socket = null;
-                    this.checkEditorReady(socket, response.editorID, response.editorUrl, action.source.editorPanel, action.source.editorActivity, outputConsole, false);
+                    this.checkEditorReady(response.editorID, response.editorUrl, action.source.editorPanel, action.source.editorActivity, outputConsole, false);
                     
 
                 } else if (responseDiagram != undefined) {
@@ -705,7 +704,6 @@ class EducationPlatformApp {
 
     /**
      * Open a websockets connection to receive status updates on editor build. 
-     * @param {Object} socket - WebSocket object
      * @param {String} statusUrl - the url for checking the status of the editor panel.
      * @param {String} editorInstanceUrl - the editor instance's url. 
      * @param {String} editorPanelId - the id of the editor panel.
@@ -713,8 +711,8 @@ class EducationPlatformApp {
      * @param {Panel} logPanel - the panel to log progress to.
      * @param {Bool} editorReady - whether editor has finished building or not.
      */
-    checkEditorReady(socket, editorID, editorInstanceUrl, editorPanelId, editorActivityId, logPanel, editorReady){
-        socket = new WebSocket(this.wsUri);
+    checkEditorReady(editorID, editorInstanceUrl, editorPanelId, editorActivityId, logPanel, editorReady){
+        var socket = new WebSocket(this.wsUri);
         socket.onopen = function(){
             socket.send(editorID);
         };
@@ -738,14 +736,14 @@ class EducationPlatformApp {
             //reconnect now
             if (!editorReady){
                 if(!socket || socket.readyState == 3){
-                    this.checkEditorReady(socket, editorID, editorInstanceUrl, editorPanelId, editorActivityId, logPanel, editorReady);
+                    this.checkEditorReady(editorID, editorInstanceUrl, editorPanelId, editorActivityId, logPanel, editorReady);
                 }
             }
         }.bind(this);
 
             
         if(!socket || socket.readyState == 3){
-            this.checkEditorReady(socket, editorID, editorInstanceUrl, editorPanelId, editorActivityId, logPanel, editorReady);
+            this.checkEditorReady(editorID, editorInstanceUrl, editorPanelId, editorActivityId, logPanel, editorReady);
         }
     }
 }
--- a/platform/src/EducationPlatformApp.js
+++ b/platform/src/EducationPlatformApp.js
@@ -463,7 +463,7 @@ class EducationPlatformApp {
                     // Language workbench
                     PlaygroundUtility.longNotification("Building editor");
 
-                    this.checkEditorReady(response.editorID, response.editorUrl, action.source.editorPanel, action.source.editorActivity, outputConsole, false);
+                    this.checkEditorReady(response.editorID, response.editorUrl, action.source.editorPanel, action.source.editorActivity, outputConsole);
                     
 
                 } else if (responseDiagram != undefined) {
@@ -709,10 +709,10 @@ class EducationPlatformApp {
      * @param {String} editorPanelId - the id of the editor panel.
      * @param {String} editorActivityId - TODO remove as this can be found using editorPanelId to save having to specify in config.
      * @param {Panel} logPanel - the panel to log progress to.
-     * @param {Bool} editorReady - whether editor has finished building or not.
      */
-    checkEditorReady(editorID, editorInstanceUrl, editorPanelId, editorActivityId, logPanel, editorReady){
+    checkEditorReady(editorID, editorInstanceUrl, editorPanelId, editorActivityId, logPanel){
         var socket = new WebSocket(this.wsUri);
+        var editorReady = false;
         socket.onopen = function(){
             socket.send(editorID);
         };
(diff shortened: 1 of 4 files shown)


platform/src/EducationPlatformApp.js
```diff
@@ -460,7 +462,9 @@ class EducationPlatformApp {
                 if (response.editorUrl) {
                     // Language workbench
                     PlaygroundUtility.longNotification("Building editor");
-                    this.checkEditorReady( response.editorStatusUrl, response.editorUrl, action.source.editorPanel, action.source.editorActivity, outputConsole);
+
+                    var socket = null;
+                    this.checkEditorReady(socket, response.editorID, response.editorUrl, action.source.editorPanel, action.source.editorActivity, outputConsole, false);
```
szschaler, line 467: Why is the socket parameter needed? As far as I can tell, it's always treated just like a local variable...

platform/src/EducationPlatformApp.js
```diff
@@ -700,48 +704,48 @@ class EducationPlatformApp {
     }
 
     /**
-     * Poll for editor to become available. 
+     * Open a websockets connection to receive status updates on editor build. 
+     * @param {Object} socket - WebSocket object
```
szschaler, line 708: See above. Explain why this parameter is needed or remove it.

platform/src/EducationPlatformApp.js
```diff
@@ -700,48 +704,48 @@ class EducationPlatformApp {
...
      * @param {String} statusUrl - the url for checking the status of the editor panel.
      * @param {String} editorInstanceUrl - the editor instance's url. 
      * @param {String} editorPanelId - the id of the editor panel.
      * @param {String} editorActivityId - TODO remove as this can be found using editorPanelId to save having to specify in config.
      * @param {Panel} logPanel - the panel to log progress to.
+     * @param {Bool} editorReady - whether editor has finished building or not.
```
szschaler, line 714: Ditto. Why is this parameter needed?

platform/src/EducationPlatformApp.js
```diff
@@ -700,48 +704,48 @@ class EducationPlatformApp {
...
+    checkEditorReady(socket, editorID, editorInstanceUrl, editorPanelId, editorActivityId, logPanel, editorReady){
+        socket = new WebSocket(this.wsUri);
+        socket.onopen = function(){
+            socket.send(editorID);
+        };
+
+        socket.onmessage = function(e){
+            var resultData = JSON.parse(e.data);
+            if (resultData.output){
+                    logPanel.setValue(resultData.output);
```
szschaler, line 725: Would be better to append output to avoid the console always jumping to the very end of the text even when nothing has actually changed.

This would require the server sending increments rather than full text. 

I would suggest that we move this into a separate issue, however, so that we can merge the core WebSocket functionality before improving its behaviour.

platform/src/EducationPlatformApp.js
```diff
@@ -700,48 +704,48 @@ class EducationPlatformApp {
...
+                }
```
szschaler, line 726: Indentation error?

platform/src/EducationPlatformApp.js
```diff
@@ -700,48 +704,48 @@ class EducationPlatformApp {
...
+            if(resultData.editorReady){
+                editorReady = true;
```
szschaler, line 728: If I understand correctly, then the only reason this is here is to stop the `onclose()` handler from trying to reconnect the websocket. This doesn't need `editorReady` to be a parameter to the function, a local variable in `checkEditorReady()` would do, too, wouldn't it? That would keep the interface cleaner.

platform/src/EducationPlatformApp.js
```diff
@@ -700,48 +704,48 @@ class EducationPlatformApp {
...
+                socket.close();
                 sessionStorage.setItem( editorPanelId , editorInstanceUrl );
                 this.activityManager.setActivityVisibility(editorActivityId, true);
                 Metro.notify.killAll();
                 PlaygroundUtility.successNotification("Building complete.");
             }
+        }.bind(this);
```
szschaler, line 735: Why do you need to do this binding here but not for the `onopen` handler? (Probably just my lack of understanding of how JS closures work...)

platform/src/EducationPlatformApp.js
```diff
@@ -700,48 +704,48 @@ class EducationPlatformApp {
...
 
-        } else {
-            console.log("ERROR: The editor response could not be checked: " + statusUrl);
-            PlaygroundUtility.errorNotification("Failed to start the editor.");
+        socket.onclose = function(){
+            //reconnect now
```
szschaler, line 738: Add some more explanation of why this is needed and how it is intended to work.

platform/src/EducationPlatformApp.js
```diff
@@ -700,48 +704,48 @@ class EducationPlatformApp {
...
+            if (!editorReady){
+                if(!socket || socket.readyState == 3){
+                    this.checkEditorReady(socket, editorID, editorInstanceUrl, editorPanelId, editorActivityId, logPanel, editorReady);
+                }
+            }
+        }.bind(this);
+
+            
+        if(!socket || socket.readyState == 3){
```
szschaler, line 747: This leads to a stack overflow if the WS server cannot be reached for some reason.

platform/src/EducationPlatformApp.js
```diff
@@ -700,48 +704,48 @@ class EducationPlatformApp {
...
+        }.bind(this);
 
-        } else {
-            console.log("ERROR: The editor response could not be checked: " + statusUrl);
-            PlaygroundUtility.errorNotification("Failed to start the editor.");
+        socket.onclose = function(){
+            //reconnect now
+            if (!editorReady){
+                if(!socket || socket.readyState == 3){
+                    this.checkEditorReady(socket, editorID, editorInstanceUrl, editorPanelId, editorActivityId, logPanel, editorReady);
```
szschaler, line 741: I suspect this might lead to some sort of overflow error if the WS server suddenly becomes unavailable. We're not cleaning up the socket we've just created and we are constantly recursively calling `checkEditorReady()`...

szschaler: Also, when would `socket` ever be `NULL`?

platform/src/EducationPlatformApp.js
szschaler: Also, when would `socket` ever be `NULL`?

szschaler (changes requested):
Added some questions and thoughts.

In addition, it would be important to document this protocol somewhere suitable in the Wiki.

platform/src/EducationPlatformApp.js
aryaei2000: This is fixed now. Socket is no longer an input parameter.

platform/src/EducationPlatformApp.js
aryaei2000: Removed now. Used it like socket when I had a slightly different implementation and was supposed to act as a guard for recursion. But now it is defined locally in the function.

platform/src/EducationPlatformApp.js
aryaei2000: This clause in the conditional can be removed.

aryaei2000: Regarding stack overflow, yes it is possible. Quickest fix for it is to develop the `onerror` case where we can guard recursion with the help of a flag. One scenario which we should add is checking every couple of seconds or every time page is refreshed for web socket server availability.

platform/src/EducationPlatformApp.js
aryaei2000: This won't fall into a never ending recursion because we have the `editorReady` as a guard. If the server doesn't set `editorReady` at all, then the client keeps creating new connections but not at the speed of light! It has to wait for a connection to get dropped before a new one is created, which is usually a matter of a few seconds.

platform/src/EducationPlatformApp.js
szschaler: Sounds good to me

platform/src/EducationPlatformApp.js
szschaler: I don't understand this argument. If the server goes down, then `editorReady` won't change value and `socket` will always have `readyState == 3` (because without a server, there's no connection that can be established). So this will just recurse away quickly and lead to a stack overflow.

szschaler (approved):

PR #222, opened 2025-02-26
220 feedback survey url

### LINKED ISSUE #220: FEEDBACK_SURVEY_URL Missing in Code
FEEDBACK_SURVEY_URL is defined in the platform Dockerfile below:

https://github.com/mdenet/educationplatform/blob/d655a97813c6c37bf3dc2582556a77d292a0cd78/platform/Dockerfile#L22

However, it is not read from an environment file and is not defined as a global constant directive to the bundler. Therefore, the value doesn't propagate to the container.

aryaei2000:
Fixes #220

### DIFF
--- a/platform/Dockerfile
+++ b/platform/Dockerfile
@@ -15,6 +15,7 @@ COPY platform/ ./platform/
 # Refer to readme.md for a complete list of supported environment varibles
 ARG TOKEN_SERVER_URL
 ARG WEBSOCKETS_URI
+ARG FEEDBACK_SURVEY_UR
 
 RUN npm ci --workspace=platform\
         && npm run build --workspace=platform --\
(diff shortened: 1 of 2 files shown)


platform/Dockerfile
```diff
@@ -15,6 +15,7 @@ COPY platform/ ./platform/
 # Refer to readme.md for a complete list of supported environment varibles
 ARG TOKEN_SERVER_URL
 ARG WEBSOCKETS_URI
+ARG FEEDBACK_SURVEY_UR
```
szschaler, line 18:
```suggestion
ARG FEEDBACK_SURVEY_URL
```

szschaler (changes requested):
I think there's a small typo ;-)

szschaler (approved):

PR #223, opened 2025-03-03
setError() isn't a function on consoles

szschaler:
This corrects a typo in the code.


aryaei2000 (approved):
Tested and looking good on dev

Issue #164, opened 2024-01-16
Branch names that include '/' separators fail to load


barnettwilliam:
Loading and saving activities from branch names that include '/' separators is not supported. The platform incorrectly parses the url therefore requests an invalid url.

Support for handling branches with '/'s should be added to [platform/src/FileHandler.js](https://github.com/mdenet/educationplatform/blob/main/platform/src/FileHandler.js) 

> Note support for saving to branches should be added - https://github.com/mdenet/educationplatform/issues/24

Issue #150, opened 2023-12-11
Visual feedback when there are saveable changes


barnettwilliam:
The save button should be disabled and the icon greyed out when all panels are in a clean state i.e. there are no modifications to the panels contents to save, it should be enabled when a panel is in a dirty state - as suggested https://github.com/mdenet/educationplatform/pull/148#pullrequestreview-1774921175

szschaler:
I think this may actually be tricky with ACE, as it doesn't seem to expose an easy event for this. We can use editor-level `change` events to _enable_ undo buttons, but the undo manager doesn't expose any events on reset or mark clean...

Issue #141, opened 2023-12-07
Add editor panel undo button


barnettwilliam:
There is no undo button for editor panels. Feedback from the MDENet Symposium indicated that an undo button would be helpuful so that any changes made can be easily reverted and would reassure learners who are unfamiliar with the models/languages used when modifying the panel contents.

jgsuess:
This requires command pattern and sessions, I am unsure if these are part of the implementation. You would need a framework that does this from the start.

agarciadom:
This may just be part of the Ace editor or even the browser's basic functionality. If I just go to the [Ace website](https://ace.c9.io/) and type something into their demo editor, I can then press Ctrl+Z for undo. I assume this functionality will be available through some browser/Ace editor API.

szschaler:
Yes, ACE has an `UndoManager` that will make this trivial to add. It's just about making available existing functionality also through a button.

Issue #140, opened 2023-12-07
Single commit save


barnettwilliam:

szschaler:
@barnettwilliam, can you add a little bit more description, please?

barnettwilliam:
If multi-file commits aren't supported by the GitHub API then this can be achieved using a separate branch and squash merging the changes.

barnettwilliam:
> @barnettwilliam, can you add a little bit more description, please?

Added

Also related to #24

szschaler:
You can use the GitHub database API to construct this step-by-step: https://siddharthav.medium.com/push-multiple-files-under-a-single-commit-through-github-api-f1a5b0b283ae

PR #227, opened 2025-04-05
Refining the Git integration in the MDENET Education Platform

hasnainn19:
# Overview

The platform’s Git integration was enhanced with several new features—to name a few: more reliable saving, support for branching operations such as switching, creating, and merging branches, and the ability to review outstanding changes.

---

## Related Issues / Discussions

During this work, several open issues were addressed:  
- #150
- #141
- #140 
- #139 
- #135 
- #25 
- #24
- #164 

---

## Changes made to the codebase

- Refactored `initialize()` in `EducationPlatformApp` such that it does not prompt for a login on each page reload.
- Refactored the panel class structure to separate panels that can be saved (`ProgramPanel`, `xTextEditorPanel`) from panels that are used for informational purposes (`ConsolePanel`, `OutputPanel`).
- Refactored the tokenserver file structure to improve organization and enforce a clearer separation of concerns, including the introduction of a dedicated `routes` and `config` directory.
- Added a test suite in the backend using Jasmine (the frontend uses Jasmine coupled with Karma). 
- Refactored the `FileHandler` to implement the Strategy pattern, allowing the version control provider to be selected at runtime. This lays the foundation for supporting multiple VCS integrations. Key additions include `BaseVcsProvider` and `GitHubProvider`.
- Made several additions to `EducationPlatformApp`, `Utility`, `LoginController`, `StorageController`, and `FileHandler` to support more robust saving and branching operations within the platform.
- Updated `index.html` and `custom.css` to visually support and present the above functionality in a clear and presentable manner.
- Extensive unit tests were written for the files mentioned.

---

## Testing & Validation

- [x] Manually tested learner workflows using several activities. 
- [x] Frontend and backend (tokenserver) unit tests added, all which pass.

---

## Additional Notes

- Regarding issue #164: While the platform enforces compliant branch names when branches are created through the UI, this restriction can be bypassed by creating branches directly via the GitHub interface.

- In certain activities, ProgramPanels are used to embed content from tools—for example, an XText activity may include a file path like: `"file": "{{ID-panel-turtles}}/xtext-resources/generated/meta-model.ecore"`. This causes an issue during the save process, as the platform currently treats all ProgramPanels as files to be saved, even when they serve only as inline content for a tool and don’t correspond to actual files in the repository. To resolve this, I propose introducing a distinct panel type—possibly named ToolPanel—to clearly differentiate tool-based content from saveable program files.

- The index.html file has grown increasingly complex due to the inclusion of UI elements for saving, switching branches, creating branches, merging branches, and reviewing changes. I propose breaking this file into smaller, modular components to improve maintainability and readability.

---

## Merge Readiness Checklist

- [x] Changes rebased and up-to-date with `upstream/main`  
- [x] No merge conflicts present  

---

**Author**: @hasnainn19
**Date**: 05/04/2025


szschaler:
Thanks, @hasnainn19. I have activated the running of our automated test suite, but this reports errors. Can you check what's going on here, please and try to correct (or report back here if you need help with this)?

hasnainn19:
The automated tests from the workflow now pass 

Small note: 

build-mac and build-win workflows used npm test --workspaces -- --browsers SafariNative. This worked previously in the absence of a backend test suite; however due to now adding a test suite in the tokenserver, this was running the following commands in each workspace:

platform: `karma start --browsers SafariNative`
tokenserver: `jasmine --config=test/support/jasmine.json --browsers SafariNative`

The `browsers` parameter is supported by Karma, which is used for running browser-based frontend tests, but the backend only uses Jasmine, so passing `browsers` into the tokenserver test script was causing it to crash. 

Therefore, I separated the test commands in `build-mac` and `build-win` as the following:

`npm test --workspace=platform -- --browsers SafariNative`
`npm test --workspace=tokenserver`

and likewise for `EdgeHeadless`.

On the other hand, running `npm test --workspaces` works as normal as the browsers are chosen in the karma configuration file rather than manually passing in a browser through the CLI.

szschaler:
Thanks. Still seems to be breaking when run from the pull request, but I suspect this is actually to do with the rights we give PR-run checks rather than with your code. I had made some fixes here before, believing these would deal with the issue (we've seen this in relation to dependabot PRs before), but clearly didn't get this right yet... @aryaei2000 could you take a look?

szschaler:
@hasnainn19 many thanks for this PR. I now have a better setup, where checks are actually run for PRs, so have finally been able to run the unit tests for your code. On the face of it, it looks good, but I notice that some errors are thrown and then caught during the unit tests. Would you be able to take a quick look to see if these come from your code or from something else?

platform/src/SaveablePanel.js
```diff
@@ -0,0 +1,134 @@
...
+    getFileUrl() {
+        return this.fileUrl;
+    }
+
+    setFileUrl(url) {
+        this.fileUrl = url;
+    }
+
+    /**
+     * Currently only supports GitHub URLs.
```
szschaler:
```suggestion
     * FIXME: Currently only supports GitHub URLs.
```
I guess we would want to generalise this at some point (though not right now...)

platform/src/EducationPlatformApp.js
```diff
@@ -144,7 +180,62 @@ class EducationPlatformApp {
...
+        });
+
+        document.addEventListener("keydown", function(event) {
+            // Check for Ctrl+S or Command+S (for macOS)
+            if ((event.ctrlKey || event.metaKey) && event.key.toLowerCase() === "s") {
+
+                // Prevent the default save action to the webpage
+                event.preventDefault();
+
+                // Can use this to trigger savePanelContents()
```
szschaler:
```suggestion
                // TODO: Can use this to trigger savePanelContents()
```

platform/src/FileHandler.js
```diff
@@ -1,30 +1,42 @@
...
-    base64ToBytes(base64) {
-        const binString = window.atob(base64);
-        return Uint8Array.from(binString, (m) => m.codePointAt(0));
-    }
-
-    bytesToBase64(bytes) {
-        const binString =  String.fromCodePoint(...bytes);
-        return window.btoa(binString);
+        if (!provider) {
+            throw new Error(`Host URL '${parsedUrl.host}' is not supported.`);
```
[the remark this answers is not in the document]
szschaler: Will we need to introduce a generic URL "provider"?

platform/test/spec/testPanelSpec.js
```diff
@@ -0,0 +1,198 @@
...
+
+    describe("visibility", () => {
+        it("can be set to invisible", () => {
+            panel.setVisible(false);
+            expect(panel.isVisible()).toBe(false);
+        });
+
+        it("handles non-boolean values", () => {
+            panel.setVisible(null);
+            expect(panel.isVisible()).toBe(null);
```
szschaler: Why is this useful behaviour?

szschaler (commented):
@hasnainn19, many thanks for this PR. I have some questions. Would you be able to help me understand better? Otherwise, this looks quite good subject to me trying to run it :-)

hasnainn19:
> @hasnainn19 many thanks for this PR. I now have a better setup, where checks are actually run for PRs, so have finally been able to run the unit tests for your code. On the face of it, it looks good, but I notice that some errors are thrown and then caught during the unit tests. Would you be able to take a quick look to see if these come from your code or from something else?

Hi, I'm not quite sure what you mean by this...

I see that all the tests have passed - unless you are talking about the excessive ERROR logs when running the unit tests, which come from the application's existing error-handling logic, for example, when the code calls `console.error` when an operation fails. A number of unit tests intentionally mock failure responses to verify that these cases are handled correctly.

On the other hand, for the tests in the tokenserver, there are no logs as the jasmine helper file [tokenserver/test/helpers/silenceLogs.js](https://github.com/mdenet/educationplatform/pull/227/files#diff-9064bb5e5c699d31ac2f272a6b588699e20629422659aee34d01596367f21594) has stubbed all the logs so they don't appear in the console output (they can still be referenced and called as usual). Although this wasn't reflected in the educationplatform test suite, as there were existing test cases that used console output assertions, and I didn't want to break their implementation.

platform/src/index.js
```diff
@@ -18,9 +18,10 @@ window.fit = platform.fit.bind(platform);
...
 window.panels = platform.panels;
-window.savePanelContents = platform.savePanelContents.bind(platform);
+window.showSaveConfirmation = platform.showSaveConfirmation.bind(platform);
+window.showBranches = platform.showBranches.bind(platform);
+window.reviewChanges = platform.reviewChanges.bind(platform);
 window.toggle = platform.toggle.bind(platform);
 window.togglePanelById = platform.togglePanelById.bind(platform);
 //window.renderDiagram = renderDiagram;
-window.longNotification = PlaygroundUtility.longNotification;
-window.getPanelTitle = platform.getPanelTitle.bind(platform);
```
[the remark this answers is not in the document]
hasnainn19: I'm not sure why it says `window.longNotification = PlaygroundUtility.longNotification;` was removed when it wasn't.

I removed `window.getPanelTitle = platform.getPanelTitle.bind(platform);` as I moved the title to be within the [Panel.js](https://github.com/mdenet/educationplatform/pull/227/files#diff-5e4f7485ea85ae741a087b1cc4ee50d4bf83f9db7c2500b87ab59b3305033cfa) class.

platform/src/OutputPanel.js
```diff
@@ -1,16 +1,16 @@
 /*global Metro ace -- Metro and ace are externally imported*/
-import { ProgramPanel } from "./ProgramPanel.js";
+import { Panel } from "./Panel.js";
 import { Button } from "./Button.js";
 import svgPanZoom from 'svg-pan-zoom';
-class OutputPanel extends ProgramPanel {
+class OutputPanel extends Panel {
```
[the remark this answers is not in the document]
hasnainn19: I don't believe so. 

Syntax highlighting is set when `this.editor.session.setMode()` applies an ACE mode for a given language. Both ProgramPanel and OutputPanel handle their syntax highlighting logic separately. 

ProgramPanel only gets its highlighting when EducationPlatformApp explicitly calls `setEditorMode(language)` during panel creation [here](https://github.com/mdenet/educationplatform/pull/227/files#diff-0a83eeac868ff1ebd0c13a448e95313199c3f0b8489b1c22bb6794a42909d54fR424).

OutputPanel applies all of its syntax highlighting internally in `initialize()`, in the “highlight” dialog, and when switching between generated files, and never calls or relies on `ProgramPanel.setEditorMode()`, which was unused. This, along with some further refactoring, showed that OutputPanel didn't need to inherit from ProgramPanel, and so I cut the connection and simply made OutputPanel inherit from Panel instead. 

However, there is also some syntax highlighting logic in XTextEditorPanel (setting the editor mode to an ACE mode for a given language), which seems to just be a duplicate of what's going on in ProgramPanel, and this is probably something that can be refactored further.

platform/src/SaveablePanel.js
hasnainn19: I think this is something that can be moved alongside the Strategy pattern with the Filehandler and VCS provider files. Instead of doing `panel.getFilePath()`, we could add a method in the FileHandler called `getFilePath(fileUrl)`. This would lead to `findProvider`, which would choose the VCS provider at runtime, and each VCS provider can have its own implementation of finding a file path from a given URL.

platform/src/FileHandler.js
```diff
@@ -56,151 +69,180 @@ class FileHandler {
...
+                }
             }
-            
-            responsePromise = jsonRequest( request.url,  JSON.stringify(request.params), true );
 
-        } else {
-            console.log("File could not be stored - not authenticated.");
-            responsePromise = null;
+            // At this point, this is either a public repository, or it's a private repository but an unknown type of URL.
+            // In either case, we assume that we can simply access the URL directly.
```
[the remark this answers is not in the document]
hasnainn19: `findProvider` throws an error only when the URL's host is not supported by any provider. It does not throw for private/public URLs in general, and it won't throw if the host is supported. 

This case from the code snippet refers to when the URL is supported by a provider; however, it's a private URL (or at least the `isPrivate` flag has been passed as true in the function argument).

platform/src/index.js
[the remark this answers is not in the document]
szschaler: Weird. I can indeed see the code in your branch, but the PR insists it has been removed...

platform/src/EducationPlatformApp.js
```diff
@@ -72,70 +80,98 @@ class EducationPlatformApp {
...
-
-        if (!urlParamPrivateRepo()){
-            // Public repo so no need to authenticate
-            this.initializeActivity(urlParameters);
-            
-        } else {
-            PlaygroundUtility.showLogin();
+        document.getElementById("btnnologin").onclick = async () => {
+            utility.setAuthenticated(false);
+            PlaygroundUtility.hideLogin();
```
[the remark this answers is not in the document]
szschaler: As per above comments...

hasnainn19:
> > I've started doing some manual testing on this. The first thing I hit is that if I choose to "continue without login", I seem to run into a loop, where the login window is shown again rather than allowing me to progress to the activity. At least that's what happens when I patiently step through in debug mode. When I just let the code run, I simply get an empty screen, but the login dialogue is removed and no activity is shown. This is when loading one of the default activities, which come from a public repository...
> > Not quite sure why this is happening. Is is something you see at your end, too, @hasnainn19 ?
> 
> I think this is because [here](https://github.com/hasnainn19/educationplatform/blob/4e197be108295acb555a87db38a272445a510b90/platform/src/EducationPlatformApp.js#L92-L95) we're not also calling `initializeActivity()`. We do this for the authenticated state (specifically, [here](https://github.com/hasnainn19/educationplatform/blob/4e197be108295acb555a87db38a272445a510b90/platform/src/EducationPlatformApp.js#L190)), but not for the unauthenticated state...


This is essentially two problems in one:

1) 

Looking through the code and the diffs, I believe that while I was refactoring `initialize()` in EducationPlatformApp, I ended up [removing the functionality that allows us to load activities for public repos](https://github.com/mdenet/educationplatform/pull/227/files#diff-0a83eeac868ff1ebd0c13a448e95313199c3f0b8489b1c22bb6794a42909d54fL81-L86), and forgot to add it back. The code checked if the activity comes from a public repository, and if so, then we can simply skip all authentication logic and head straight to the activity. We should probably add this back, as right now, if the activity comes from a public repository, we have to go through the authentication phase, which is unnecessary at this point. 

2)

If we choose "continue without login" (which should only ever happen if we are trying to load an activity from a **private** repository, since if it's public, we shouldn't even see this button), I kept the functionality the same as it was previously, which as you can see [here](https://github.com/mdenet/educationplatform/blob/main/platform/src/EducationPlatformApp.js#L75), didn't do anything except hide the login. This is because if we want to access a private repository, we need to authenticate with GitHub, and if we don't, there's not much we can do. Perhaps a "continue without logging in" option shouldn't even be available since it doesn't lead anywhere.

szschaler (changes requested):
Suggestion for fixing the panel-visibility tests

platform/test/spec/testPanelSpec.js
szschaler: Great. I have added a code-change suggestion...

szschaler:
> > > I've started doing some manual testing on this. The first thing I hit is that if I choose to "continue without login", I seem to run into a loop, where the login window is shown again rather than allowing me to progress to the activity. At least that's what happens when I patiently step through in debug mode. When I just let the code run, I simply get an empty screen, but the login dialogue is removed and no activity is shown. This is when loading one of the default activities, which come from a public repository...
> > > Not quite sure why this is happening. Is is something you see at your end, too, @hasnainn19 ?
> > 
> > 
> > I think this is because [here](https://github.com/hasnainn19/educationplatform/blob/4e197be108295acb555a87db38a272445a510b90/platform/src/EducationPlatformApp.js#L92-L95) we're not also calling `initializeActivity()`. We do this for the authenticated state (specifically, [here](https://github.com/hasnainn19/educationplatform/blob/4e197be108295acb555a87db38a272445a510b90/platform/src/EducationPlatformApp.js#L190)), but not for the unauthenticated state...
> 
> This is essentially two problems in one:
> 
> 
> Looking through the code and the diffs, I believe that while I was refactoring `initialize()` in EducationPlatformApp, I ended up [removing the functionality that allows us to load activities for public repos](https://github.com/mdenet/educationplatform/pull/227/files#diff-0a83eeac868ff1ebd0c13a448e95313199c3f0b8489b1c22bb6794a42909d54fL81-L86), and forgot to add it back. The code checked if the activity comes from a public repository, and if so, then we can simply skip all authentication logic and head straight to the activity. We should probably add this back, as right now, if the activity comes from a public repository, we have to go through the authentication phase, which is unnecessary at this point.
> 
> If we do add this back, we should probably add a `setupUnauthenticatedState()` function, similar to [setupAuthenticatedState()](https://github.com/hasnainn19/educationplatform/blob/main/platform/src/EducationPlatformApp.js#L197) but without the stuff that comes from being authenticated.
> 
> Although this logic doesn't support the case when the activity is in a public repository, the user is marked as a contributor to it, and wants to make changes and save them. In this case, they would need to be authenticated so they can have access to authenticated actions (saving, branching, merging, etc). However this depends on whether the platform should support modifying and saving changes from activities in public repositories which I believe at the moment does not.
> 
> 
> If we choose "continue without login" (which should only ever happen if we are trying to load an activity from a **private** repository, since if it's public, we shouldn't even see this button), I kept the functionality the same as it was previously, which as you can see [here](https://github.com/mdenet/educationplatform/blob/main/platform/src/EducationPlatformApp.js#L75), didn't do anything except hide the login. This is because if we want to access a private repository, we need to authenticate with GitHub, and if we don't, there's not much we can do. Perhaps a "continue without logging in" option shouldn't even be available since it doesn't lead anywhere.

Thanks, @hasnainn19, for looking into this. I did notice that you hadn't removed the `privaterepo` URL parameter and that we now needed to make a log-in decision even for public repositories :-) I've added code that reacts correctly when no-login is chosen and that works OK now. I think I prefer to keep this in, because it allows the option to enable saving even for public repositories (assuming the user has write access to those repositories, which they can only prove by logging in with GitHub). It is  bit frustrating, however, to have to make an explicit log-in decision for the public examples... 

WDYT?

szschaler:
> > In testing this on my local machine, I've identified an issue with swallowing errors due to invalid files. I _thought_ we had unit tests to check for this, but clearly these aren't working correctly... See comments below...
> 
> I think this is a mistake on my part, where I overengineered the error handling and put `initializeActivity()` in a try-catch block when in fact, the errors from `initializeActivity()` were already being displayed. In this case, I think we can revert to the original state and remove the try-catch block surrounding `initializeActivity()`, and potentially, `setupAuthenticatedState()` in general.
> 
> If I recall correctly, I think I separated initializeActivity into its own try-catch block in the first place because the errors shown on the display weren't detailed enough, and I wanted a better view of what was going wrong so I directed it to a `console.error` instead and forgot to change it back. Perhaps in the implementation of initializeActivity we can embed some error logs so that a developer can check the console for more in-depth logs if something goes wrong.

I like this, @hasnainn19. Will just removing the try-catch-block be enough (given that these calls are deep in some promises)? I like the idea of logging the errors to the console, that can be easily added at the end of `initializeActivity()`...

hasnainn19:
> It is bit frustrating, however, to have to make an explicit log-in decision for the public examples...

As a suggestion for a better UX:

In that case, why not skip the login process for public examples, and add an option for logging in (potentially as a Sign In button in the corner of the page) when already at the activity? This makes it so that they don't have to explicitly choose whether to log in or not, start off as unauthenticated, and can choose to log in afterwards if they wish to have access to more actions (saving, branching, etc). Once authenticated, the visibility of this button would be turned off.

platform/src/FileHandler.js
```diff
@@ -1,30 +1,42 @@
...
-        const binString =  String.fromCodePoint(...bytes);
-        return window.btoa(binString);
+        if (!provider) {
+            throw new Error(`Host URL '${parsedUrl.host}' is not supported.`);
+        }
+        return provider;
     }
 
+    /**
+     * TODO: Remove this function and use the one below - this makes it async, so we need to also go back and update the callers.
```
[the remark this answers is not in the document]
hasnainn19: Rather, we would need to modify the chain of function calls to propagate the async/await context, and stop until we reach an entry point (i.e `initialize`), an event handler, or an already async context. 

Below is the asynchronous propagation chain:

FileHandler.fetchFile()
  │
  ├── ActivityManager.fetchActivities()
  │   └── ActivityManager.initializeActivities()
  │       └── EducationPlatformApp.initializeActivity()
  │           ├── EducationPlatformApp.setupAuthenticatedState()
  │           │   └── EducationPlatformApp.handleAuthRedirect() **[already async with try/catch]**
  │           │
  │           └── EducationPlatformApp.initialize() **[entry point - top-level initialization]**
  │
  ├── ActivityManager.fetchFile()
  │   └── ActivityManager.resolveRef()
  │       └── ActivityManager.fetchActivity()
  │           └── EducationPlatformApp.initializePanels()
  │               └── EducationPlatformApp.initializeActivity() **[already changed above]**
  │
  └── EducationPlatformApp.isLocalEnvironmentOutdated()
      └── EducationPlatformApp.savePanelContents() **[already async]**

platform/src/FileHandler.js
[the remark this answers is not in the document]
szschaler: Thanks for the explanation. This makes sense.

szschaler:
> > Would you have a minute to make these changes?
> 
> I can try to mock something up, and then we can review it to see what more may need to be done.

That would be brilliant!

hasnainn19:
@szschaler 

> * [x]  For a private repository, there should not be an option to not log in. The login modal should show as the first thing the user sees.
> * [x]  For a public repository, the login modal should not show by default, but there should be a sign-in button somewhere in the left-hand menu, in case they want to save their state. 
> 

I've also added an "Account" button in the sidebar so that users can see the account they are logged in with, and have an option to either sign in (if they are on an activity while unauthenticated, which can only happen if they're on a public activity) or sign out, which will clear all authentication data and return them to the login modal. 

A future improvement would be to analyze the permissions of the GitHub token on the given repository. From this we can see if the user has read access (if not, currently initializeActivity just displays an error saying it couldn't fetch the activity) or write access. If there is no write access then we can disable certain actions such as saving, creating branches, merging branches, etc.

szschaler (dismissed):
This now looks great and ready to merge! Many thanks for all your work on this. I will merge as soon as I have resolved some issues with testing on staging server and get this deployed onto the production platform ASAP

hasnainn19:
> Just noticed this:
> 
> > In certain activities, `ProgramPanels` are used to embed content from tools—for example, an XText activity may include a file path like: `"file": "{{ID-panel-turtles}}/xtext-resources/generated/meta-model.ecore"`. This causes an issue during the save process, as the platform currently treats all `ProgramPanels` as files to be saved, even when they serve only as inline content for a tool and don’t correspond to actual files in the repository. To resolve this, I propose introducing a distinct panel type—possibly named `ToolPanel`—to clearly differentiate tool-based content from saveable program files.
> 
> Is this indeed still a problem? In the original implementation this was fixed by simply ignoring content URLs that were full URLs rather than relative to the location of the activity file. Can we not continue to do this?

Can I ask what you mean by this? From my understanding, each SaveablePanel has a `fileUrl` field, which corresponds to the `raw.githubusercontent` url of that file in the repository, e.g `https://raw.githubusercontent.com/user/repo/main/library.ocl`. Meanwhile, some other ProgramPanels would have the fileUrl `{{ID-panel-turtles}}/xtext-resources/generated/meta-model.ecore` where `{{ID-panel-turtles}}` would be interpolated to a full URL at runtime.

szschaler:
If memory serves, then the original save method would check the URL and only attempt to save if it was a GitHub URL. I think that internally we stored the exact ref used in the activity specification (modulo any `{{ID-...}}` rewriting), and would actually only save if the ref wasn't a full URL but a relative path, because then we know it's a file in the same repository.

platform/src/SaveablePanel.js
hasnainn19: > I think this is something that can be moved alongside the Strategy pattern with the Filehandler and VCS provider files. Instead of doing `panel.getFilePath()`, we could add a method in the FileHandler called `getFilePath(fileUrl)`. This would lead to `findProvider`, which would choose the VCS provider at runtime, and each VCS provider can have its own implementation of finding a file path from a given URL.

Implemented in [9573744](https://github.com/mdenet/educationplatform/pull/227/commits/957374455390776c68e187bccb3a47dbfa5ce336)

Each VCS provider has its own structure for a raw URL, so we delegate this to its respective provider class.

szschaler:
> The problem I experienced was that when loading the [XText activity](https://github.com/mdenet/educationplatform-examples/blob/main/xtext-turtles/xtext_activity.json), the `panel-mm` panel with the name "XMI Text Format":
> 
> ```
> {
>     "id": "panel-mm",
>     "name": "XMI text format",
>     "ref": "ecore",
>     "file": "{{ID-panel-turtles}}/xtext-resources/generated/meta-model.ecore"
> },
> ```
> 
> image
> 
> is visible in the layout, of class ProgramPanel, and can be edited (should this be read-only information?). Therefore if we edit it then `canSave` will return true and we will try to save it, returning an error as explained before.

Yes, I had thought about this, too. I agree in that case we would get an error. But this is a separate issue (it already occurs in the current platform implementation). Your PR introduces a new issue where these panels don't work at all...

platform/src/EducationPlatformApp.js
```diff
@@ -859,7 +859,7 @@ class EducationPlatformApp {
         const panelsToSave = this.getPanelsWithChanges();
         for (const panel of panelsToSave) {
 
-            const remoteFile = await this.fileHandler.fetchFile(panel.getFileUrl(), utility.urlParamPrivateRepo());
+            const remoteFile = await this.fileHandler.fetchFileFromRepository(panel.getFileUrl(), utility.urlParamPrivateRepo());
             if (!remoteFile) {
                 throw new Error(`No remote file found for ${panel.getTitle()}`);
```
szschaler: If we skipped this file rather than throwing an exception, we could handle non-repo URLs more easily. We would then just skip them in the actual `save` method, too....

platform/src/FileHandler.js
```diff
@@ -101,12 +125,35 @@ export class FileHandler {
 
             // At this point, this is either a public repository, or it's a private repository but an unknown type of URL.
             // In either case, we assume that we can simply access the URL directly.
-            const responseText = await utility.getRequest(url, false);
-            return { content: responseText, sha: null };
+            const response = await this.refactoredFetchFileDirectly(url);
+            if (response != null) {
+                return { content: response.content, sha: null }; // TODO need to retrieve the sha for the file
```
szschaler: There probably isn't one, so `null` seems like a sensible choice.

platform/src/FileHandler.js
```diff
@@ -101,12 +125,35 @@ export class FileHandler {
...
+     * @returns {Promise<{content: string} | null>} Promise resolving to object with content, or null if failed
+     */
+    async refactoredFetchFileDirectly(url) {
+        try {
+            const responseText = await utility.getRequest(url, false);
+            return { content: responseText };
         } 
         catch (error) {
-            console.error("Failed to fetch file:", error);
+            console.error("Failed to fetch file with url: " + url, error);
```
szschaler: Should this be allowed to float through to the platform to be shown as an error to the user?

platform/src/EducationPlatformApp.js
hasnainn19: To stop attempting to save non-repo panels, how about just excluding them in `canSave()`? 

```diff
  /**
  * Has the contents of this panel changed since the last save?
  * @returns {boolean} true if the content has changed since the last save
  */
 canSave() {
-    return super.getValue() !== this.getLastSavedContent();
+    // Exclude panels that are from outside the repository
+    if (!this.getFileUrl().includes("raw.githubusercontent.com")) {
+        return false;
+    }
+
+    return super.getValue() !== this.getLastSavedContent();
 }
```


Although if in the future we rely on loading panel content from an external public repository, this would fail. We are also hard-coding the reliance on GitHub here, but it doesn't seem wise to provide a `FileHandler` to a `SaveablePanel`.

platform/src/FileHandler.js
hasnainn19: Since we're now allowing users to commit changes to public repositories, all the panels loaded from one will have `null` SHAs which is an issue...

platform/src/FileHandler.js
```diff
@@ -101,12 +125,35 @@ export class FileHandler {
...
-            return { content: responseText, sha: null };
+            const response = await this.refactoredFetchFileDirectly(url);
+            if (response != null) {
+                return { content: response.content, sha: null }; // TODO need to retrieve the sha for the file
+            }
+            return null;
+
+        } 
+        catch (error) {
+            console.error("Failed to fetch file with url: " + url, error);
```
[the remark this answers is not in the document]
hasnainn19: This happens when a URL from outside the repository cannot be fetched. Do you propose displaying this error in a `PlaygroundUtility` notification?

platform/src/FileHandler.js
szschaler: Ah, I see. I had missed that. How difficult is it to grab the SHA here?

platform/src/EducationPlatformApp.js
hasnainn19: Implemented in [a2b41ab](https://github.com/mdenet/educationplatform/pull/227/commits/a2b41aba57beb4dc08f5d9b44618181455f18f85)

platform/src/FileHandler.js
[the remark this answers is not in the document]
hasnainn19: I would like to avoid simply introducing a `PlaygroundUtility` object to the FileHandler and keep the classes decoupled, and continue down the established methods of handling errors in the codebase, such as using `PlaygroundUtility` or `EducationPlatformError` in the cases of loading an activity. If a file could not be fetched, the convention is to return `null` and the caller should therefore have a null-check. 

[591afa4](https://github.com/mdenet/educationplatform/pull/227/commits/591afa4e38735436ebba5fb2ce83b19565fe814b) provides the foundation for this approach. This should also work with the refactored versions of the methods, which return null if they did not get a success response. Essentially, the job of showing these types of errors to the user is for the callers in their null-check.

szschaler:
I can confirm that loading from a non-repo file for the contents of a panel works.

platform/src/ActivityManager.js
```diff
@@ -126,28 +126,27 @@ class ActivityManager {
...
-                                                    Check the activity file is available at the given url and you have the correct access rights.`) );
+        } 
+        else {
+            errors.push(new EducationPlatformError(`
+                The activity configuration file was not accessible at: ${this.activitiesUrl}. 
+                Check the activity file is available at the given url and you have the correct access rights.
+            `));
         }
 
-        if (fileContent != null){
```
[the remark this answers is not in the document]
hasnainn19: Just above we perform if `(file && file.content) then fileContent = file.content` to ensure that we didn't get a null, so at this point `fileContent` is non-null.

szschaler:
This is now running on the [staging server](https://mdenet-dev.sites.er.kcl.ac.uk). @hasnainn19 , can you have a quick play with it there? You will need this [GitHub App](https://github.com/apps/mdenet-app-for-staging-server) to be able to sign in.

From a quick trial, I cannot see any issues... (though staging can be a bit slow...) (Note that for some reason there's currently an issue with the staging server, but I don't think that's to do with your code and will hopefully go away when you get around to trying this...)

